PR #3, opened 2017-02-17
Update composer.json to allow PHP 7.0 installs

### LINKED ISSUE #2: SDK 0.2 does not support PHP 7 via Composer
Due to php's version constrict in composer.json, the MyParcel SDK v0.2.x will only install on PHP 5.6.5 and later minor releases, but will not install on PHP 7.0 and above.

Is there a reason PHP 7.0 and above are excluded?

LeonidasJP:
Fixes #2 by allowing Composer.json to install PHP 7.0 as well

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "url": "https://github.com/myparcelnl/sdk/issues"
     },
     "require": {
-        "php": "~5.6.5"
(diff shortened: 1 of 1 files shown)


reindert-vetter:
Thanks for your contribution

Issue #6, opened 2017-02-22
Impossible to add parcels without Customer Email


LeonidasJP:
When the E-mail address of the customer is not known, it is not possible to send a parcel through the SDK; API replies "E-mail should not be null".

In some cases, customers have not provided an e-mail address with their order, but the parcel should still be shipped through MyParcel. Eg. Phone sale or Counter sale with no direct pickup possibilities.

Is it possible to default MyParcelConsignment::email to <empty_string> instead of null, so these shipments can be accepted by the API? Or otherwise write an extra check on adding a Consignment to MyParcelCollection whether the E-mail address is still null?

PR #8, opened 2017-02-22
Add consignment without E-mail information

LeonidasJP:
This PR allows SDK users to push a consignment, if there is no E-mail address available, by defaulting to an empty string, instead of null.
Implements #6


reindert-vetter:
Thanks!

PR #34, opened 2017-12-20
Fixed setting the pick up fields. 

rhamoudi:
Data being returned by the api no longer uses pickup_postal_code but postal_code instead.


RichardPerdaan:
Hi @rhamoudi,

Thank you for noticing this, I have accepted the pull request.

PR #75, opened 2018-10-04
Fix unit test send one international consignment

RichardPerdaan:


src/Model/Repository/MyParcelConsignmentRepository.php
```diff
@@ -666,32 +668,52 @@ private function decodeExtraOptions($data)
 
         if (key_exists('company', $recipient)) {
             $this->setCompany($recipient['company']);
+        } else {
+            $this->setCompany(false);
```
reindert-vetter: false is of type boolean, but the function expects a string.

src/Model/Repository/MyParcelConsignmentRepository.php
```diff
@@ -666,32 +668,52 @@ private function decodeExtraOptions($data)
...
         }
 
-        if (key_exists('only_recipient', $recipient)) {
-            $this->setOnlyRecipient($recipient['only_recipient']);
+        if (key_exists('number', $recipient)) {
+            $this->setNumber($recipient['number']);
+        } else {
+            $this->setNumber(false);
```
reindert-vetter: false is of type boolean, but the function expects a integer.

reindert-vetter: Zie ook andere functies

Issue #80, opened 2018-10-08
Links made with setLinkOfLabels() always return A4 pdfs


EdieLemoine:
`getRequestBody()` isn't added to the request in `setLinkOfLabels()` rendering it unable to return A6 pdf links

PR #81, opened 2018-10-08
#80 links made with setlinkoflabels() always return a4 pdfs

EdieLemoine:
- rename setA4() to setPosition() because it can can create A6 as well depending on its argument
- send getRequestBody() with setLinkOfLabels() allowing it to create A6 pdfs. add constant DEFAULT_A4_POSITION = 1 as default value for $positions argument so it still creates A4 as default (useful for customers)

### DIFF
--- a/src/Helper/MyParcelCollection.php
+++ b/src/Helper/MyParcelCollection.php
@@ -367,7 +367,7 @@ class MyParcelCollection
         /** If $positions is not false, set paper size to A4 */
         $this
             ->createConcepts()
-            ->setPosition($positions);
+            ->setLabelFormat($positions);
 
         $conceptIds = $this->getConsignmentIds($key);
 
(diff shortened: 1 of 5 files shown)


src/Helper/MyParcelCollection.php
```diff
@@ -352,7 +353,7 @@ public function setLatestDataWithoutIds($key, $size = 300)
     /**
      * Get link of labels
      *
-     * @param array|int|bool $positions The position of the label on an A4 sheet. You can specify multiple positions by
+     * @param array|int|bool $positions The position of the label on an A4 sheet. Set to false to create an A6 sheet. You can specify multiple positions by
      *                                  using an array. E.g. [2,3,4]. If you do not specify an array, but specify a
      *                                  number, the following labels will fill the ascending positions. Positioning is
      *                                  only applied on the first page with labels. All subsequent pages will use the
```
reindert-vetter: Php-docs heeft soft warping. Je moet je phpstorm afstemmen op 120.

src/Helper/MyParcelCollection.php
```diff
@@ -550,7 +551,7 @@ private function getConsignmentReferenceIds(&$key)
      *
      * @return $this
      */
-    private function setA4($positions = 1)
```
reindert-vetter: Rename `setLabelFormat()`

PR #88, opened 2018-10-19
+ Added street additional info setter

bastiaansenglenn:
I didn't find any setter for additional info field like; 'third room'. I created a setter for a variabele and in the get method i checked if it's set then use the custom variabele.


bastiaansenglenn:
Sorry for the previous pull request, didn't get it working right

EdieLemoine:
We modified your code and it's now in PR #89

PR #107, opened 2019-02-08
retun the first consignment when reference_identifier is not set

RichardPerdaan:


src/Helper/MyParcelCollection.php
```diff
@@ -102,6 +102,10 @@ public function getOneConsignment()
      */
     public function getConsignmentsByReferenceId($id)
     {
+        if ($this->count() === 1) {
```
reindert-vetter:
```suggestion
        if ($this->count() === 1 && $id === null) {
```

PR #110, opened 2019-02-26
add Exception for getConsignmentsByReferenceId()

RichardPerdaan:


src/Helper/MyParcelCollection.php
```diff
@@ -99,10 +99,15 @@ public function getOneConsignment()
      * @param string|null $id
      *
      * @return MyParcelCollection
+     * @throws \Exception
      */
     public function getConsignmentsByReferenceId($id)
     {
-        if ($this->count() === 1 && $id === null) {
+        if ($id === null) {
+            throw new \Exception('Can\'t run getConsignmentsByReferenceId(): id can\'t be null');
```
reindert-vetter: referenceId can\'t be null

RichardPerdaan: Done!

PR #114, opened 2019-04-08
fix-support-collection-class 

Mvzundert:
Operator for where and operatorForWhere should be NULL by default according to the Laravel 5.7 documentation (https://laravel.com/api/5.7/Illuminate/Support/Collection.html)

It generated and error when trying to install or use the SDK package because of the mismatch.

### DIFF
--- a/src/Support/Collection.php
+++ b/src/Support/Collection.php
@@ -515,7 +515,7 @@ class Collection extends CollectionProxy implements ArrayAccess, Countable, Iter
      * @param  mixed  $value
      * @return static
      */
-    public function where($key, $operator, $value = null)
+    public function where($key, $operator = null, $value = null)
     {
         return $this->filter($this->operatorForWhere(...func_get_args()));
     }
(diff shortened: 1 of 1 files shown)


reindert-vetter:
Thank you very much!

Mvzundert:
@reindert-vetter This seems to have disappeared from the current master version (so it breaks again for Laravel 5.7, 5.8)

RichardPerdaan:
Hi @Mvzundert, 
When I look at the master branch on lines [518](https://github.com/myparcelnl/sdk/blob/master/src/Support/Collection.php#L518) and [531](https://github.com/myparcelnl/sdk/blob/master/src/Support/Collection.php#L531) I see the adjustment. I think it has nothing to do with this change.

I see that you have opened a new [pull request](https://github.com/myparcelnl/sdk/pull/117/) with a different modification. Thank you for investigating and we will look into this.

Mvzundert:
@RichardPerdaan I wanted to update this comment as I found the issue myself :) so yes it's a Laravel 5.8 support issue and has nothing to do with this PR.

RichardPerdaan:
@Mvzundert I understand from this that the problem has been solved. Thanks again for your adjustment.

PR #174, opened 2019-10-29
Update ConsignmentEncode.php to fix Invoice on Customs Declaration

coderiekelt:
After being baffled as to why my customs declarations were displaying my label description as the invoice number, I found this little sneaky fella which I think may be the wrong getter being called in the encoder.


RichardPerdaan:
Thank you for reporting that it is indeed not correct. We are going to adjust it.

PR #197, opened 2019-12-24
Fix for Uncaught TypeError: strlen() expects parameter 1 to be string

magefan:
Fatal error: Uncaught TypeError: strlen() expects parameter 1 to be string, null given in vendor/myparcelnl/sdk/src/Model/Consignment/AbstractConsignment.php:703


RichardPerdaan:
This problem has been solved via the following pull request #202.
Thank you for thinking along, for the solution.

PR #206, opened 2020-02-04
use getItemValue

RichardPerdaan:


src/Model/Consignment/BpostConsignment.php
```diff
@@ -214,7 +214,7 @@ public function validate(): bool
         /** @var \MyParcelNL\Sdk\src\Model\MyParcelCustomsItem $item */
         $amount = 0;
```
jasperpoppe: PP: Maybe change the name of the variable to `$totalValue`? The term "amount" is used as "total items" in this case.

RichardPerdaan: Done

PR #212, opened 2020-02-26
use public const for INSURANCE_POSSIBILITIES_LOCAL

RichardPerdaan:


src/Model/Consignment/BpostConsignment.php
```diff
@@ -6,12 +6,29 @@
...
     public const DEFAULT_WEIGHT = 50;
-    public const INSURANCE_500 = 500;
-    public const INSURANCE_0   = 0;
-    public const CARRIER_ID    = 2;
-    public const CARRIER_NAME  = 'bpost';
 
+    /**
+     * @var array
+     */
+    public const INSURANCE_POSSIBILITIES_LOCAL = [0, 500];
```
RichardPerdaan:
```suggestion
    public const INSURANCE_POSSIBILITIES_LOCAL = [0, 500, 1000, 1500, 2000];
```

PR #230, opened 2020-05-14
Compatible with Laravel 7

renalpha:
Fixes:
PHP Fatal error:  Declaration of MyParcelNL\Sdk\src\Support\Collection::when($value, callable $callback, ?callable $default = NULL) must be compatible with Illuminate\Support\Collection::when($value, ?callable $callback = NULL, ?callable $default = NULL) in /vendor/myparcelnl/sdk/src/Support/Collection.php on line 38

### DIFF
--- a/src/Support/Collection.php
+++ b/src/Support/Collection.php
@@ -479,11 +479,11 @@ class Collection extends CollectionProxy implements ArrayAccess, Countable, Iter
      * Apply the callback if the value is truthy.
      *
      * @param  bool  $value
-     * @param  callable  $callback
-     * @param  callable  $default
+     * @param  callable|null  $callback
+     * @param  callable|null  $default
      * @return mixed
      */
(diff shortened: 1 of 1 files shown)


Cannonb4ll:
Is it possible to get this merged? My application update is being held back because of this issue.

RickJeroen:
This is already fixed in #225?

renalpha:
> This is already fixed in #225?

yes

Cannonb4ll:
I still seem to have this issue, not sure if you guys tried this but it has not been solved for me. Using version 4.

Cannonb4ll:
Just checked, if you use version 4 the collection is not updated properly:

https://github.com/myparcelnl/sdk/blob/ce167d0f2ea76a8506114bafdac65a67683cce57/src/Support/Collection.php#L486

RichardPerdaan:
On this moment we work on v4.1.0. I expect to set this version live today

PR #236, opened 2020-06-02
fixed returning of false for method apiKeyIsCorrect in CheckApiKeySer…

jordye:
Currently when the API key is not valid you will encounter the following error:
``strpos() expects parameter 1 to be string, object given``

Instead of passing the `Excpetion` object the actual message is used now, which I believe was intended initially.


RichardPerdaan:
Hi @jordye ,
the branch is out of date with the develop branch, can you update your branch.  I can't do that for you.

PR #238, opened 2020-06-05
it should be possible to use capitalized abbreviations

RichardPerdaan:


reindert-vetter:
This has been fixed in develop and will be included with the next 5.x release.

PR #276, opened 2020-11-10
fix fatal on missing destination country regex

yaroslav-zenin:
Fix
Fatal Error: 'Uncaught TypeError: preg_match() expects parameter 1 to be string, null given in /..../vendor/myparcelnl/sdk/src/Helper/ValidateStreet.php:39

### DIFF
--- a/src/Helper/ValidateStreet.php
+++ b/src/Helper/ValidateStreet.php
@@ -36,7 +36,12 @@ class ValidateStreet
      */
     public static function validate(string $fullStreet, string $localCountry, ?string $destinationCountry): bool
     {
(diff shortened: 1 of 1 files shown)


reindert-vetter:
The following PR will also solve this problem:
https://github.com/myparcelnl/sdk/pull/275/files

@RichardPerdaan @robindelaater do you know when it can be merged?

RichardPerdaan:
Hi @yaroslav-zenin,
Would you like to check if version 5.2.3 of the MyParcel SDK works for you?

yaroslav-zenin:
> Hi @yaroslav-zenin,
> Would you like to check if version 5.2.3 of the MyParcel SDK works for you?

Hi
yes, 5.2.3 fixed that problem
thanks

PR #277, opened 2020-11-16
Fix receiving shopId, fix error when null values are allowed

blieb:
This pull requests fixes setting shop_id to consignment when receiving it from api

it also fixes when values of a consignment are NULL. it did already show in comments that NULL was allowed, but not in declaration of function so it did give errors when it was NULL.

### DIFF
--- a/src/Model/Consignment/AbstractConsignment.php
+++ b/src/Model/Consignment/AbstractConsignment.php
@@ -586,7 +586,7 @@ class AbstractConsignment
     /**
      * @return integer|null
      */
-    public function getShopId(): ?int
+    public function getShopId(): int
     {
         return $this->shop_id;
     }
--- a/src/Model/Consignment/AbstractConsignment.php
+++ b/src/Model/Consignment/AbstractConsignment.php
@@ -586,7 +586,7 @@ class AbstractConsignment
     /**
      * @return integer|null
      */
-    public function getShopId(): int
+    public function getShopId(): ?int
     {
         return $this->shop_id;
     }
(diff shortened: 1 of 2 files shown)


src/Model/Consignment/AbstractConsignment.php
```diff
@@ -586,7 +586,7 @@ public function setStatus($status): self
     /**
      * @return integer|null
      */
-    public function getShopId(): int
+    public function getShopId(): ?int
```
reindert-vetter, line 589: @RichardPerdaan in @blieb situation, the shop id is present. Do you think we should allow shop id null or not? If we don't allow it, we'll be more likely to notice that something unexpected is happening (like the above bug in ConsignmentAdapter.php). For getPickupStreet, getPickupCity and getPickupNumber it is valid to return null.

reindert-vetter (changes requested):

src/Model/Consignment/AbstractConsignment.php
```diff
@@ -586,7 +586,7 @@ public function setStatus($status): self
     /**
      * @return integer|null
```
reindert-vetter, line 587:
```suggestion
     * @return integer
```

src/Model/Consignment/AbstractConsignment.php
reindert-vetter, line 589:
```suggestion
    public function getShopId(): int
```

reindert-vetter (changes requested):

reindert-vetter (approved):

PR #280, opened 2020-12-11
Fix multiline streets not getting parsed properly

tdgroot:
We were receiving the following error:

```
An error has occurred while validating order number <curated>. Check address.-MyParcelNL\Sdk\src\Exception\InvalidConsignmentException: Something went wrong splitting up the following address: Some Street
123
5 A
```

### DIFF
--- a/src/Helper/SplitStreet.php
+++ b/src/Helper/SplitStreet.php
@@ -77,6 +77,8 @@ class SplitStreet
             foreach (self::NUMBER_SUFFIX_ABBREVIATION as $from => $to) {
                 $fullStreet = preg_replace("/(\d.*-?)[\s]$from/", '$1' . $to, $fullStreet);
             }
+
+            $fullStreet = trim(preg_replace('/(\r\n)|\n|\r/', ' ', $fullStreet));
         }
 
         if ($destination === AbstractConsignment::CC_BE) {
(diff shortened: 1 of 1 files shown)


src/Helper/SplitStreet.php
```diff
@@ -77,6 +77,8 @@ public static function splitStreet(string $fullStreet, string $local, string $de
             foreach (self::NUMBER_SUFFIX_ABBREVIATION as $from => $to) {
                 $fullStreet = preg_replace("/(\d.*-?)[\s]$from/", '$1' . $to, $fullStreet);
             }
+
+            $fullStreet = trim(preg_replace('/(\r\n)|\n|\r/', ' ', $fullStreet));
```
reindert-vetter, line 81: I think it's good to add a test to the split street tests.
Or you can of put this logic in a function. Than you can make a unit test for that function.

reindert-vetter (changes requested):

src/Helper/SplitStreet.php
reindert-vetter, line 81: I see the same thing happening on line 92. You can take that away. If you took away that line, I will accept your pull-request.

reindert-vetter (changes requested):

src/Helper/SplitStreet.php
tdgroot: Duplicate code removed.

reindert-vetter:
I think the adjustment could be simpler. I have created a pull request: https://github.com/myparcelnl/sdk/pull/284/files
Would you like to check this PR?

reindert-vetter:
Fixed in https://github.com/myparcelnl/sdk/pull/284/files

PR #302, opened 2021-04-26
My 28141 show friendlier error messages

joerivanveen:
By actually validating some basic data it is easier to prevent bulk errors through the api, in addition any error thrown by the api is parsed especially to try to get the actual offending order_id to feedback to the client.


joerivanveen:
This is now fixed in the plugin itself.

PR #310, opened 2021-07-16
fix: use method to get the request url

RichardPerdaan:
If you want to connect the plugins to a staging, it is possible to adjust the API_BASE_URL (in the plugin). Only an adjustment has to be made in the sdk to actually get the label from staging.

### DIFF
--- a/src/Helper/MyParcelCollection.php
+++ b/src/Helper/MyParcelCollection.php
@@ -425,7 +425,7 @@ class MyParcelCollection extends Collection
                 )
                 ->sendRequest('GET', $requestType);
 
-            $this->label_link = (new MyParcelRequest())->getRequestUrl(). $request->getResult("data.$urlLocation.url");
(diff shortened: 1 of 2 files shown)


src/Helper/MyParcelCollection.php
```diff
@@ -425,7 +425,7 @@ public function setLinkOfLabels($positions = self::DEFAULT_A4_POSITION): self
                 )
                 ->sendRequest('GET', $requestType);
 
-            $this->label_link = MyParcelRequest::REQUEST_URL . $request->getResult("data.$urlLocation.url");
+            $this->label_link = (new MyParcelRequest())->getRequestUrl(). $request->getResult("data.$urlLocation.url");
```
joerivanveen, line 428: Spatie voor de punt...

RichardPerdaan: Done

EdieLemoine (approved):

joerivanveen (approved):

PR #378, opened 2022-01-06
refactor(php8): remove deprecation warnings

rowhel95:
Forgot these

### DIFF
--- a/src/Support/Collection.php
+++ b/src/Support/Collection.php
@@ -1778,7 +1778,7 @@ class Collection implements ArrayAccess, Countable, IteratorAggregate
      *
      * @return \ArrayIterator
      */
-    public function getIterator(): Traversable
+    public function getIterator(): ArrayIterator
     {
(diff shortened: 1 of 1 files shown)


src/Support/Collection.php
```diff
@@ -1778,7 +1778,7 @@ public function toJson($options = 0)
      *
      * @return \ArrayIterator
      */
-    public function getIterator()
+    public function getIterator(): Traversable
```
EdieLemoine, line 1781: Why not `ArrayIterator`?

EdieLemoine (changes requested):

src/Support/Collection.php
rowhel95: Because I'm not sure if the return type needs an exact match or not

EdieLemoine: If you change it to `ArrayIterator` I'll approve the PR

it-can:
Any news on this?

PR #436, opened 2023-01-08
fix(consignment): type error on getting shop id

Jamie4224:
When the shop_id is null in AbstractConsignment, the getShopId() method would not function because it was typed as int and not as ?int, meaning a null return was not allowed. This has been fixed in this PR.


EdieLemoine:
Hi Jamie, if you could rebase your pull request onto our `main` branch, we can merge this change.

Jamie4224:
Hi! I think it should be good now.

PR #492, opened 2024-07-22
fix: prevent adding array to int calculating total value

joerivanveen:
INT-589

### DIFF
--- a/test/Rule/MinimumItemValueRuleTest.php
+++ b/test/Rule/MinimumItemValueRuleTest.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Sdk\Test\Rule;
+
+use MyParcelNL\Sdk\src\Model\Consignment\PostNLConsignment;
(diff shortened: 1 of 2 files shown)


GravendeelJochem (approved):

test/Rule/MinimumItemValueRuleTest.php
```diff
@@ -0,0 +1,79 @@
...
+     * @return array
+     */
+    public function provideMinimumItemValueData(): array
+    {
+        return [
+            [
+                [
+                    'items' => [
+                        [
+                            'amount' => 50,
```
FlorianSDV, line 24:
```suggestion
                            'value' => 50,
```

test/Rule/MinimumItemValueRuleTest.php
```diff
@@ -0,0 +1,79 @@
...
+                        ],
+                        [
+                            'amount' => 50,
+                        ],
+                    ],
+                ],
+                true,
```
FlorianSDV, line 31:
```suggestion
                'expectation' => true,
```

PR #517, opened 2025-01-06
feat: allow sender with feature header

joerivanveen:
Also fixes the bug where multishop would not work.

INT-781
INT-804

### DIFF
--- a/src/Model/Consignment/AbstractConsignment.php
+++ b/src/Model/Consignment/AbstractConsignment.php
@@ -18,6 +18,7 @@ use MyParcelNL\Sdk\src\Helper\ValidatePostalCode;
 use MyParcelNL\Sdk\src\Model\Carrier\AbstractCarrier;
 use MyParcelNL\Sdk\src\Model\Carrier\CarrierFactory;
 use MyParcelNL\Sdk\src\Model\MyParcelCustomsItem;
+use MyParcelNL\Sdk\src\Model\Recipient;
 use MyParcelNL\Sdk\src\Services\CountryCodes;
 use MyParcelNL\Sdk\src\Services\CountryService;
 use MyParcelNL\Sdk\src\Support\Str;
@@ -520,7 +521,9 @@ abstract class AbstractConsignment
     /**
      * @var null|string
      */
-    private $state;
+    private           $state;
+
+    private Recipient $sender;
 
     /**
      * @throws \Exception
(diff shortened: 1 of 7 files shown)


GravendeelJochem (approved):

src/Model/Consignment/AbstractConsignment.php
```diff
@@ -522,6 +523,8 @@ abstract class AbstractConsignment
      */
     private $state;
 
+    private $sender;
```
FreekVR, line 526: As we do not yet require PHP 7.4 for the SDK, we can't use property types yet, but could you at least type it with a PHPDoc annotation?

src/Model/Consignment/AbstractConsignment.php
```diff
@@ -532,6 +535,34 @@ public function __construct()
...
+    public function setSender(Recipient $sender): AbstractConsignment
+    {
+        $this->sender = $sender;
+
+        return $this;
+    }
+
+    public function hasSender(): bool
+    {
+        return isset($this->sender) && Recipient::class === get_class($this->sender);
```
FreekVR, line 554: Judging by the implementation of `setSender` and the lack of default - whenever `$this->sender` has a value, it should be the `Recipient`-class. So this seems to be a bit redundant?

src/Model/Consignment/AbstractConsignment.php
```diff
@@ -532,6 +535,34 @@ public function __construct()
...
+    }
+
+    public function getSender(): ?Recipient
```
FreekVR, line 557: This could then also just straight up return `$this->sender` as it doesn't seem possible to have any other value than `null` or an instance of Recipient.

FreekVR (changes requested):

src/Model/Consignment/AbstractConsignment.php
joerivanveen: While it is true what you say now, I have experienced that later additions to the code can make this false. I will remove it for now, but generally I like to be thorough and not cause errors unnecessarily down the line. (When hasSender is checked and the object turns out to be something else afterall.)

src/Model/Consignment/AbstractConsignment.php
joerivanveen: Using isset is cleaner (and I think maybe also faster?) than just returning something and have it convert to a boolean in the process?

FreekVR: When it's strongly typed, this isn't needed - but as that isn't currently possible this solution is acceptable at the moment.

FreekVR (approved):

PR #533, opened 2025-06-27
fix(php-sdk): correctly pass attributes for each multi-collo shipment

GravendeelJochem:
INT-731


FreekVR (approved):

src/Helper/MyParcelCollection.php
```diff
@@ -172,11 +172,26 @@ public function addConsignment(AbstractConsignment $consignment): self
         if ($consignment->getApiKey() === null) {
             throw new MissingFieldException('First set the API key with setApiKey() before running addConsignment()');
         }
-
         $consignment->validate();
-
         $this->push($consignment);
 
+        // Check if all consignments have the same carrier
```
joerivanveen, line 178: The collection may contain consignments for different carriers and even consignments with different api keys.

src/Helper/MyParcelCollection.php
```diff
@@ -172,11 +172,26 @@ public function addConsignment(AbstractConsignment $consignment): self
...
+        $carrierIds = array_map(function($c) { return $c->getCarrierId(); }, $this->items);
+        if (count(array_unique($carrierIds)) > 1) {
+            throw new \Exception('All consignments in a multi collo shipment must have the same carrier.');
+        }
+
+        // Only set multi collo if all reference_identifiers are the same (or not set)
```
joerivanveen, line 184: I understand the urge to ‘automate’ it here, but
- now the addMultiCollo method has gone, this is a breaking change that we would like to prevent
- a consuming application should (imo) have the choice to add consignments as multi collo or not (like the current situation)
- the check whether the carrier has multi collo ability is missing I think

joerivanveen: My understanding from the talk was that
- you refactor the addDataList (or something like that) to a method that accepts a collection itself and can set this collection as a multi collo consignment as such (with different weights and options)
- keep the addMultiCollo(consignment, amount) and realy it to your new method (with weight equally distributed) for backwards compatibility

joerivanveen (changes requested):

src/Helper/MyParcelCollection.php
```diff
@@ -172,14 +172,70 @@ public function addConsignment(AbstractConsignment $consignment): self
...
+     * @throws \Exception
+     */
+    public function addMultiColloConsignments(array $consignments): self
+    {
+        if (empty($consignments)) {
+            return $this;
+        }
+
+        // Check if all consignments have the same carrier
+        $carrierIds = array_map(function($c) { return $c->getCarrierId(); }, $consignments);
```
joerivanveen, line 223:
```suggestion
        $carrierIds = array_map(static function($consignment) {
            return $consignment->getCarrierId();
        }, $consignments);
```

src/Helper/MyParcelCollection.php
```diff
@@ -172,14 +172,70 @@ public function addConsignment(AbstractConsignment $consignment): self
...
+        if (count(array_unique($carrierIds)) > 1) {
```
joerivanveen, line 224: Yoda style is preferred but not a reason to dismiss this pr.

joerivanveen (approved):

PR #538, opened 2025-07-15
fix!: use correct ups constants

joerivanveen:
This also updates the authors of the sdk to current maintainers and removes unused carrier_id and carrier_name constants from the consignments, superseded by the property carrierClass.


FreekVR (approved):
looks great!

PR #545, opened 2025-08-08
test(mock): introduce Mockery (HTTP mock) for SDK tests

NabDevs:
This PR migrates the entire test suite to a standardized Mockery-based HTTP mocking system. This modernizes our testing infrastructure and provides more flexible, maintainable test mocking capabilities.

- Use Mockery with the standardized mockCurl() method
- Updated base TestCase class to provide consistent mocking interface
- Docs: test/Mock/README.md with Mockery usage examples and test/Mock/EXAMPLE_UPDATE_TEST.md migration guide
- Goal: no real (mutating) API calls in tests

Resolves INT-1079

### DIFF
--- a/src/Helper/MyParcelCurl.php
+++ b/src/Helper/MyParcelCurl.php
@@ -127,7 +127,6 @@ class MyParcelCurl
             return $this->handleHeaderLine($curl, $headerLine, $headers);
         });
 
-
         $response = curl_exec($resource);
 
         // Check the return value of curl_exec()


src/Model/MyParcelRequest.php
```diff
@@ -409,6 +414,15 @@ private function handleErrors(string $url): void
      */
     private function instantiateCurl(): MyParcelCurl
     {
+        $mockEnabled = getenv('MP_SDK_TEST');
```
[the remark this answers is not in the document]
NabDevs: Idem.

test/Bootstrap/TestCase.php
```diff
@@ -32,6 +32,24 @@ public function __construct($name = null, array $data = [], $dataName = '')
...
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+        
+        // Enable mocking if not explicitly disabled
+        if (getenv('MP_SDK_TEST') !== 'false') {
+            // Register the mock class if available
+            if (class_exists('\MyParcelNL\Sdk\Test\Mock\MockMyParcelCurl')) {
+                \MyParcelNL\Sdk\Model\MyParcelRequest::setOverride(
```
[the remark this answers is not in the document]
NabDevs: Will slack you about this.

test/Bootstrap/TestCase.php
```diff
@@ -32,6 +32,24 @@ public function __construct($name = null, array $data = [], $dataName = '')
...
 
+    /**
+     * Set up mock for all tests
+     */
+    protected function setUp(): void
+    {
+        parent::setUp();
+        
+        // Enable mocking if not explicitly disabled
+        if (getenv('MP_SDK_TEST') !== 'false') {
```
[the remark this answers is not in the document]
NabDevs: Good to know. Will use the existing env var then.

src/Helper/MyParcelCurl.php
```diff
@@ -127,6 +127,7 @@ public function getResponse(): array
             return $this->handleHeaderLine($curl, $headerLine, $headers);
         });
 
+
```
GravendeelJochem, line 130:
```suggestion
```

test/Model/Consignment/ConsignmentOtherOptionsTest.php
```diff
@@ -71,6 +107,39 @@ public function testAutoDetectPickup(array $testData): void
...
     {
+        // Mock HTTP client for API calls
+        $mockCurl = $this->mockCurl();
+        
+        // Get the appropriate response set from the dataset
+        $responses = ShipmentResponses::getPostNLFlow([
+            'reference_identifier' => $testData[self::REFERENCE_IDENTIFIER] ?? null,
+            'save_recipient_address' => $testData[self::SAVE_RECIPIENT_ADDRESS] ?? false,
+            'package_type' => $testData[self::PACKAGE_TYPE] ?? AbstractConsignment::PACKAGE_TYPE_PACKAGE,
+            'country' => $testData[self::COUNTRY] ?? 'NL',
```
joerivanveen, line 118: I would prefer it if this and similar occasions would be replaced by something more readable using a default address array (constant), e.g. (pseudo code):
```
getPostNLFLow(
$testData + self::DEFAULT_ADDRESS + [
           'reference_identifier' => $testData[self::REFERENCE_IDENTIFIER] ?? null,
            'save_recipient_address' => $testData[self::SAVE_RECIPIENT_ADDRESS] ?? false,
            'package_type' => $testData[self::PACKAGE_TYPE] ?? AbstractConsignment::PACKAGE_TYPE_PACKAGE,
            ]);
```

Or maybe more defaults could be in that constant (package type, reference identifier, etc), and the default address could be switched between countries. So maybe a little class / functionality to abstract this `getPostNLFlow(getMyValues($testData, self::DEFAULT_VALUES))`, in stead of the long winded `??` constructions in several places that seem error prone to me.

test/Mock/Datasets/ShipmentResponses.php
```diff
@@ -0,0 +1,531 @@
...
+    public static function getDHLForYouFlow(array $testData = []): array
+    {
+        return self::getStandardShipmentFlow([
+            'reference_identifier' => $testData['reference_identifier'] ?? null,
+            'carrier_id' => 9,
+            'recipient' => [
+                'cc' => $testData['country'] ?? 'NL',
+                'postal_code' => $testData['postal_code'] ?? '6825ME',
+                'city' => $testData['city'] ?? 'Arnhem',
+                'street' => 'Meander',
```
joerivanveen, line 200: If you would take a look at this file again, I think there is room for improvement, though maybe not urgently. For instance this is a Dutch street (‘Meander 631’) that will be used for other countries as well. Also I suspect there is more `??` default setting going on than necessary, although it is hard to make out exactly. In the interest of KISS I would try to confine the `??` defaulting to one place in each chain of events, it seems to me this is done multiple times now. Both things are clearly not a problem now but I fear bugs when we add things or change things. Lastly, could this be implemented as a couple of classes inheriting from an abstract response class, to keep line numbers per file low (this file is going to grow fast when adding carriers and edge cases), or is that over engineering?

test/Mock/Datasets/ShipmentResponses.php
```diff
@@ -0,0 +1,531 @@
...
+        ]);
+    }
+    
+    /**
+     * UPS specific response set
+     */
+    public static function getUPSFlow(array $testData = [], int $carrierId = 14): array
+    {
+        $country = $testData['country'] ?? 'DE';
+        $defaultStreet = $country === 'DE' ? 'Feldstrasse' : ($country === 'NL' ? 'Hoofdstraat' : 'Antareslaan');
```
joerivanveen, line 348: This is not going to scale well (what if we add a testcase for FR?) and may be fixed by adding a logic functionality / class to handle the $testData arrays as mentioned earlier.

joerivanveen (changes requested):

PR #554, opened 2025-10-14
feat: refactor platform naming to proposition for future API transition

NabDevs:
- Change Account model to use proposition_id as primary internal storage
- Add getPropositionId() method alongside deprecated getPlatformId()
- Support both platform_id and proposition_id in API responses (forward/backward compatible)
- Add setUserAgentForProposition() method alongside deprecated setUserAgent()
- Update toArray() to output proposition_id as primary key with platform_id for legacy
- Add comprehensive test coverage for all transition scenarios
- Ensure proposition_id takes precedence

This prepares the SDK for when the API team transitions from platform to proposition terminology.

INT-1087


test/Model/Account/AccountRefactoringTest.php
```diff
@@ -0,0 +1,80 @@
...
+            'shops' => []
+        ];
+
+        $account = new Account($accountDataBoth);
+
+        // proposition_id should take precedence
+        self::assertEquals(8, $account->getPlatformId(), 'Should use proposition_id when both provided');
+        self::assertEquals(8, $account->getPropositionId(), 'Should use proposition_id when both provided');
+    }
+}
```
FreekVR: missing newline here

FreekVR (changes requested):
Looks good! Smart way to keep it backwards compatible.

joerivanveen (approved):

PR #563, opened 2025-11-14
feat: add direct printing

GravendeelJochem:
INT-1245

### DIFF
--- a/src/Helper/MyParcelCollection.php
+++ b/src/Helper/MyParcelCollection.php
@@ -191,8 +191,7 @@ class MyParcelCollection extends Collection
 
         // Create multiple consignments with equally distributed weight
         $originalWeight = $consignment->getTotalWeight();
-        $weightPerCollo = $originalWeight / $amount;
-        
+        $weightPerCollo = (int) ($originalWeight / $amount);
         $consignments = [];
         for ($i = 1; $i <= $amount; $i++) {
             $clonedConsignment = clone $consignment;
@@ -225,7 +224,7 @@ class MyParcelCollection extends Collection
         }
 
         // Set multi collo and reference identifier for all consignments
-        $referenceId = $consignments[0]->getReferenceIdentifier() ?? ('multi_collo_' . uniqid('', true));
+        $referenceId = $consignments[0]->getReferenceIdentifier() ?: 'multi_collo_' . uniqid('', true);
         foreach ($consignments as $consignment) {
             $consignment->setMultiCollo(true);
             $consignment->setReferenceIdentifier($referenceId);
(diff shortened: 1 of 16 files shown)


src/Helper/MyParcelCollection.php
```diff
@@ -568,6 +569,90 @@ public function downloadPdfOfLabels($inline_download = false): void
...
+     * a special Accept header that tells the API to print directly instead of returning a PDF.
+     *
+     * @param string $printerGroupId The ID of the printer group to print to
+     *
+     * @return array The API response, grouped by API key
+     * @throws AccountNotActiveException
+     * @throws ApiException
+     * @throws MissingFieldException
+     */
+    public function printDirect(string $printerGroupId): array
```
joerivanveen: Please adapt the existing `createConsignments` to be used with this additional ‘direct print’ header and endpoint, backwards compatible if possible (it is already used for return shipments with a boolean flag, that might need to change), to prevent much duplicate code.

src/Model/MyParcelRequest.php
```diff
@@ -179,15 +199,37 @@ static function ($carry, $item) {
...
+    private $baseUrl = null;
+
     /**
      * Get request url.
      *
      * @return string
      */
     public function getRequestUrl(): string
     {
+        if ($this->baseUrl !== null) {
```
joerivanveen: Prefer yoda style.

src/Model/MyParcelRequest.php
```diff
@@ -179,15 +199,37 @@ static function ($carry, $item) {
...
+    /**
+     * Set a custom base URL for this request (e.g., for printing API)
+     *
+     * @param string $baseUrl
+     * @return self
+     */
+    public function setBaseUrl(string $baseUrl): self
+    {
+        $this->baseUrl = $baseUrl;
+        return $this;
```
joerivanveen: Prefer white line above return statement.

src/Helper/MyParcelCollection.php
```diff
@@ -568,6 +593,35 @@ public function downloadPdfOfLabels($inline_download = false): void
...
+     * directly to a printer instead of returning a PDF.
+     *
+     * @param string $printerGroupId The ID of the printer group to print to
+     *
+     * @return array The API response, grouped by API key
+     * @throws AccountNotActiveException
+     * @throws ApiException
+     * @throws MissingFieldException
+     */
+    public function printDirect(string $printerGroupId): array
```
joerivanveen: Nice job so far but the contents of this method, in an ideal world, would be `return $this->createConsignments(false, $printerGroupId);` for consistency. (And so for easier understanding and assuming in the future.)

NabDevs (approved):

src/Services/Web/PrinterGroupWebService.php
```diff
@@ -0,0 +1,45 @@
...
+     * @throws \MyParcelNL\Sdk\Exception\ApiException
+     * @throws \MyParcelNL\Sdk\Exception\MissingFieldException
+     */
+    public function getPrinterGroups(): Collection
+    {
+        $request = $this->createRequest()
+            ->setBaseUrl(MyParcelRequest::PRINTING_API_URL)
+            ->sendRequest('GET', self::ENDPOINT);
+
+        // The API may return printer groups in different structures depending on version
```
joerivanveen: I do not think this is true, and if it is, I would choose to support the current implementation and not a previous one, since this functionality is new in the SDK anyway.

joerivanveen (approved):

PR #587, opened 2026-02-13
feat: Shipment-first create flow (generated Core API client)

NabDevs:
## Short summary
Implements INT-1032 phase 2: shipment-first create flow using the generated Core API client.

## Changes
- Shipment creation now uses `ShipmentApi::postShipments()`.
- Added `ShipmentCollection` as the typed collection for shipment create/print flows.
- Added `ShipmentApiFactory` as central factory for generated Core API client setup.
- Updated `HttpCapabilitiesClient` to use `ShipmentApiFactory`.
- Updated shipment mapping/convenience code to use generated **named enum constants**.
- Direct print remains on `MyParcelRequest` because it requires a per-request custom `Accept` header (`printer-group-id`).

## Notes
- Rebases on latest `v11` (including regenerated client updates).
- Enum serialization workaround middleware has been removed after upstream OpenAPI/codegen fixes.
- Remaining temporary workaround is only for `reference_identifier` deserialization fallback in create response handling.
- Out of scope for this PR: labels, track & trace services, query/delete/returns/multi-collo.

## Validation
- `ShipmentCollectionTest`
- `ShipmentConvenienceTest`
- `ShipmentCreateServiceTest`
- `ShipmentPrintServiceTest`
- `CapabilitiesServiceTest`

INT-1032


src/Helper/ShipmentCollection.php
```diff
@@ -0,0 +1,356 @@
...
+     * for the current union schema. In that case we return null and let caller
+     * fallback to the request-side reference identifier.
+     */
+    private function normalizeReferenceIdentifier($referenceIdentifier): ?string
+    {
+        if (null === $referenceIdentifier) {
+            return null;
+        }
+
+        if (is_string($referenceIdentifier) || is_int($referenceIdentifier) || is_float($referenceIdentifier)) {
```
FreekVR: `is_scalar` might be sufficient here?

src/Model/Shipment/Shipment.php
```diff
@@ -26,6 +26,25 @@ class Shipment extends ShipmentRequest
...
+        }
+
+        $options = $this->getOptions();
+        if (is_array($options) || $options instanceof GeneratedShipmentOptionsModel) {
+            $this->setOptions($options);
+        }
+
+        if (null !== $this->getCarrier()) {
+            $this->setCarrier($this->getCarrier());
+        }
```
FreekVR: This seems to be some form of casting, but it's not really casting?

I think there are better solutions for this, for example using magic getters/setters that do not require to manually mutate data after being already present on the model instance.

src/Model/Shipment/Shipment.php
FreekVR: I would suggest strictly typing the arguments for *all* of these setters.

Either require the argument to be the same as whatever the child model accepts in its constructor. And/or allow an instance of that model itself to be passed directly.

For IDs / name-to-ID arguments, ideally we'd accept an enum of some sort, but otherwise refer to it in the docblock.

Overall, avoid mixed/untyped arguments

src/Model/Shipment/ShipmentOptions.php
```diff
@@ -0,0 +1,62 @@
...
+class ShipmentOptions extends RefShipmentShipmentOptions
+{
+    /**
+     * @param array<string, mixed>|null $data
+     */
+    public function __construct(?array $data = null)
+    {
+        parent::__construct($data);
+
+        if (null !== $this->getPackageType()) {
```
FreekVR: see my comment about casting in the Shipment model

src/Model/Shipment/PackageType.php
FreekVR: Can we not extend a generated API model here?

If we can't, this seems like it should be more like an enum of sorts rather than a model. (It seems to be an attribute of ShipmentOptions based on the code elsewhere in this PR?)

FreekVR (changes requested):

src/Collection/ShipmentCollection.php
```diff
@@ -0,0 +1,160 @@
...
+                return $shipment instanceof Shipment
+                    && $shipment->getReferenceIdentifier() === $referenceIdentifier;
+            }
+        )));
+    }
+
+    /**
+     * Filter by reference identifier prefix.
+     */
+    public function filterByReferenceIdPrefix(string $referencePrefix): self
```
FreekVR: should use `where` functionality from collections

FreekVR (changes requested):

src/Collection/ShipmentCollection.php
NabDevs: Updated. Prefix filtering now uses the collection functionality (`filter()` + `values())`. Since our `where()` implementation does not support starts-with logic, the prefix match itself lives in a filter callback.

src/Model/Shipment/Mapping/ApiMappingInterface.php
```diff
@@ -0,0 +1,22 @@
...
+    public function enumToId(string $value): int;
+
+    public function idToEnum(int $id): string;
+
+    /**
+     * @return string[]
+     */
+    public function all(): array;
+
+    public function isValid(string $value): bool;
```
FreekVR: I think this function has limited use (see docblock below regarding implementation currently)

Suggest making the existing mapping functions nullable or indicating consumers should implement them with try/catch if the intended use is validation.

```suggestion
    /**
     * Check whether the given value exists on the left-hand side of the map. Only suitable for checking v2 values.
     */
    public function isValid(string $value): bool;
```

src/Model/Shipment/Mapping/CarrierApiMapping.php
```diff
@@ -0,0 +1,78 @@
...
+        if (false === $enum) {
+            throw new InvalidArgumentException("Unknown carrier id '{$id}'");
+        }
+
+        return $enum;
+    }
+
+    public function all(): array
+    {
+        return array_keys($this->map());
```
FreekVR: I would expect this to return the whole map rather than just the keys (left-hand-side)

FreekVR (changes requested):
Nog een paar laatste dingjes, paar aanpassingen aan het contract van de mapping en wat documentatie verbeteringen.

src/Model/Shipment/Mapping/CarrierApiMapping.php
NabDevs: Agreed. all() now returns the full mapping array.

src/Model/Shipment/Mapping/CarrierApiMapping.php
```diff
@@ -0,0 +1,78 @@
...
+    public function enumToId(string $value): int
+    {
+        return (int) $this->enumToApiRef($value);
+    }
+
+    public function idToEnum(int $id): string
+    {
+        $apiRef = (string) $id;
+        $map = $this->map();
+        $enum = array_search($apiRef, $map, true);
```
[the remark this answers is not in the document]
NabDevs: Applied. idToEnum() now uses non-strict reverse lookup.

src/Model/Shipment/Mapping/PackageTypeApiMapping.php
```diff
@@ -0,0 +1,71 @@
...
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Sdk\Model\Shipment\Mapping;
+
+use InvalidArgumentException;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefShipmentPackageType;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefShipmentPackageTypeV2;
+
+final class PackageTypeApiMapping implements ApiMappingInterface
```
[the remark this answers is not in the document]
NabDevs: Done. Follows the same behaviour as `CarrierApiMapping` now.

FreekVR (approved):

joerivanveen (approved):

PR #588, opened 2026-02-16
feat: introduce label- and track & trace services for shipment domain

NabDevs:
Implements INT-1032 phase 3/4 by moving labels and track & trace flows from `MyParcelCollection` to shipment-first services.

### Changes
- Added `ShipmentLabelsService` (`src/Services/Labels/ShipmentLabelsService.php`)
- Added `ShipmentTrackTraceService` (`src/Services/TrackTrace/ShipmentTrackTraceService.php`) using generated `getTrackTracesByIds`
- Added/updated tests:
  - `test/Services/Labels/ShipmentLabelsServiceTest.php`
  - `test/Services/TrackTrace/ShipmentTrackTraceServiceTest.php`
  - `test/Services/CoreApi/ShipmentApiFactoryTest.php`
- Added live smoke coverage:
  - `test/Services/ShipmentServicesLiveSmokeTest.php`

### Notes
- Labels intentionally still use `MyParcelRequest` (generated labels flow is not sufficient yet for link/raw PDF + prepared flow).
- Track & trace uses the generated Shipments client endpoint.
- Track & trace response compatibility normalization remains centralized in `ShipmentApiFactory`.

### Test
- `ShipmentLabelsServiceTest`
- `ShipmentTrackTraceServiceTest`
- `ShipmentApiFactoryTest`
- `ShipmentServicesLiveSmokeTest`

INT-1032

### DIFF
--- a/src/Services/Labels/ShipmentLabelsService.php
+++ b/src/Services/Labels/ShipmentLabelsService.php
@@ -1,0 +1,184 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Sdk\Services\Labels;
+
+use InvalidArgumentException;
+use MyParcelNL\Sdk\Concerns\HasUserAgent;
+use MyParcelNL\Sdk\Exception\ApiException;
+use MyParcelNL\Sdk\Exception\MissingFieldException;
+use MyParcelNL\Sdk\Helper\LabelHelper;
+use MyParcelNL\Sdk\Model\MyParcelRequest;
+use setasign\Fpdi\Fpdi;
+
+final class ShipmentLabelsService
+{
+    use HasUserAgent;
+
+    private const PREFIX_PDF_FILENAME = 'myparcel-label-';
+
+    private string $apiKey;
+
+    private string $labelLink = '';
+
+    private string $labelPdf = '';
+
+    public function __construct(string $apiKey)
+    {
+        $this->apiKey = $apiKey;
+    }
+
+    /**
+     * @param int[] $shipmentIds
+     * @param mixed $positions
+     */
+    public function setLinkOfLabels(array $shipmentIds, $positions = 1): string
+    {
+        $this->validateIds($shipmentIds);
+
+        $queryString = $this->buildLabelQuery($positions);
+        $urlLocation = 'pdfs';
+        $requestType = MyParcelRequest::REQUEST_TYPE_RETRIEVE_LABEL;
+
+        if ($this->useLabelPrepare(count($shipmentIds))) {
+            $requestType = MyParcelRequest::REQUEST_TYPE_RETRIEVE_PREPARED_LABEL;
+            $urlLocation = 'pdf';
+        }
+
+        $request = (new MyParcelRequest())
+            ->setUserAgents($this->getUserAgent())
+            ->setRequestParameters(
+                $this->apiKey,
+                implode(';', $shipmentIds) . '/' . $queryString
+            )
+            ->sendRequest('GET', $requestType);
+
+        $this->labelLink = (new MyParcelRequest())->getRequestUrl() . $request->getResult("data.{$urlLocation}.url");
+
+        return $this->labelLink;
+    }
+
+    public function getLinkOfLabels(): string
+    {
+        return $this->labelLink;
+    }
+
+    /**
+     * @param int[] $shipmentIds
+     * @param mixed $positions
+     *
+     * @throws ApiException
+     */
+    public function setPdfOfLabels(array $shipmentIds, $positions = 1): string
+    {
+        $this->validateIds($shipmentIds);
+
+        $queryString = $this->buildLabelQuery($positions);
+
+        $request = (new MyParcelRequest())
+            ->setUserAgents($this->getUserAgent())
+            ->setRequestParameters(
+                $this->apiKey,
+                implode(';', $shipmentIds) . '/' . $queryString,
+                MyParcelRequest::HEADER_ACCEPT_APPLICATION_PDF
+            )
+            ->sendRequest('GET', MyParcelRequest::REQUEST_TYPE_RETRIEVE_LABEL);
+
+        $result = $request->getResult();
+
+        if (! is_string($result) || ! preg_match('/^%PDF-\d/', $result)) {
+            if (is_array($result) && isset($result['data']['payment_instructions'])) {
+                throw new ApiException('Received payment link instead of pdf. Check your MyParcel account status.');
+            }
+
+            throw new ApiException('Did not receive expected pdf response. Please contact MyParcel.');
+        }
+
+        if (! class_exists(Fpdi::class)) {
+            throw new ApiException('FPDI dependency is required to process label PDFs.');
+        }
+
+        $fpdi = new Fpdi();
+        $fileResource = fopen('php://memory', 'rb+');
+        fwrite($fileResource, $result);
+
+        $pageCount = $fpdi->setSourceFile($fileResource);
+
+        for ($i = 1; $i <= $pageCount; $i++) {
+            $templateIndex = $fpdi->importPage($i);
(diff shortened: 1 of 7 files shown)


FreekVR (changes requested):

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,184 @@
...
+
+use InvalidArgumentException;
+use MyParcelNL\Sdk\Concerns\HasUserAgent;
+use MyParcelNL\Sdk\Exception\ApiException;
+use MyParcelNL\Sdk\Exception\MissingFieldException;
+use MyParcelNL\Sdk\Helper\LabelHelper;
+use MyParcelNL\Sdk\Model\MyParcelRequest;
+use setasign\Fpdi\Fpdi;
+
+final class ShipmentLabelsService
```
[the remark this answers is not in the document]
NabDevs: Good point, thanks. For this phase I intentionally did not remove/deprecate legacy classes yet, that cleanup is planned as a separate (last) step once all shipment-first services are in place, to keep this PR scoped and reviewable.

On generated client usage: Track & Trace is now migrated to the generated CoreApi client (`getTrackTracesByIds`). Labels are still on `MyParcelRequest` for now because the generated labels flow currently does not cover the behavior we need (link extraction, raw PDF bytes, and prepared-label flow). I added this explicitly in the service docs and kept the integration point isolated so we can switch labels to generated client once that API point is complete.

src/Collection/ShipmentCollection.php
```diff
@@ -0,0 +1,160 @@
...
+ * Intended usage:
+ * - store and iterate shipments
+ * - perform in-memory filtering/lookups
+ * - no network side-effects
+ *
+ * API operations are handled by dedicated services.
+ *
+ * @property Shipment[] $items
+ */
+final class ShipmentCollection extends Collection
```
FreekVR, line 23: I think this needs to be aligned with the changes made to the other collection in your earlier PR (eg duplicated methods from the base Collection and unsuffixed names for methods)

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,193 @@
...
+     * @param int[] $shipmentIds
+     * @param mixed $positions
+     */
+    public function setLinkOfLabels(array $shipmentIds, $positions = 1): string
+    {
+        $this->validateIds($shipmentIds);
+
+        $queryString = $this->buildLabelQuery($positions);
+        $urlLocation = 'pdfs';
+        $requestType = MyParcelRequest::REQUEST_TYPE_RETRIEVE_LABEL;
```
FreekVR: I believe we can use `ShipmentApi::getShipmentsLabels()` here instead? Please see the openapi spec for the correct header (should be the default - `application/json`) to send along with the request (determines whether a link to a pdf, pdf-as-string, or an actual PDF response is given) also see https://developer.myparcel.nl/api-reference/06.shipments.html#_6-f-2

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,193 @@
...
+
+        if ($this->useLabelPrepare(count($shipmentIds))) {
+            $requestType = MyParcelRequest::REQUEST_TYPE_RETRIEVE_PREPARED_LABEL;
+            $urlLocation = 'pdf';
+        }
```
FreekVR: Should also use `getShipmentLabels()` with the header `application/vnd.shipment_label_link+json` ( > 25 orders) see https://developer.myparcel.nl/api-reference/06.shipments.html#_6-f-2

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,193 @@
...
+        $request = (new MyParcelRequest())
+            ->setUserAgents($this->getUserAgent())
+            ->setRequestParameters(
+                $this->apiKey,
+                implode(';', $shipmentIds) . '/' . $queryString,
+                MyParcelRequest::HEADER_ACCEPT_APPLICATION_PDF
+            )
+            ->sendRequest('GET', MyParcelRequest::REQUEST_TYPE_RETRIEVE_LABEL);
+
+        $result = $request->getResult();
```
FreekVR: Use ShipmentApi::getShipmentsLabels() with the correct header to get the labels as PDF

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,193 @@
...
+        if (is_numeric($positions)) {
+            return '?format=A4&positions=' . LabelHelper::getPositions((int) $positions);
+        }
+
+        if (is_array($positions)) {
+            return '?format=A4&positions=' . implode(';', $positions);
+        }
+
+        return '?format=A6';
+    }
```
FreekVR: This becomes redundant when using generated `ShipmentApi` - `getShipmentsLabels()` will take the format and positions as typed arguments and build the query internally.

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,193 @@
...
+        if (empty($ids)) {
+            throw new InvalidArgumentException('At least one shipment ID is required');
+        }
+
+        foreach ($ids as $id) {
+            if (! is_int($id) && ! (is_string($id) && ctype_digit($id))) {
+                throw new InvalidArgumentException('Shipment IDs must be integers');
+            }
+        }
+    }
```
FreekVR: This is also likely redundant as the `ShipmentApi` contains validation and is generally more type safe.

FreekVR (changes requested):
We should convert the new `ShipmentLabelsService` to use the generated API client. Otherwise, looks good.

Additionally, please aim for full test coverage.

src/Collection/ShipmentCollection.php
NabDevs: Rebased and fixed.

src/Services/Labels/ShipmentLabelsService.php
NabDevs: As dicussed, now uses the generated client.

src/Services/Labels/ShipmentLabelsService.php
NabDevs: Agreed. Fixed.

src/Services/Labels/ShipmentLabelsService.php
NabDevs: Agreed. Manual query building is removed

src/Services/Labels/ShipmentLabelsService.php
NabDevs: Removed.

src/Services/Labels/ShipmentLabelsService.php
NabDevs: Done.

FreekVR (approved):

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            throw new ApiException('Did not receive expected pdf response. Please contact MyParcel.');
+        }
+
+        if (! class_exists(Fpdi::class)) {
+            throw new ApiException('FPDI dependency is required to process label PDFs.');
+        }
+
+        $fpdi = new Fpdi();
```
joerivanveen: Fpdi is not necessary if we do just one request. You may remove it.

joerivanveen (changes requested):

src/Services/Labels/ShipmentLabelsService.php
NabDevs: Good spot. Removed it!

joerivanveen (approved):

PR #589, opened 2026-02-17
feat!: remove dropoff point from maintained code

joerivanveen:
INT-1417
Also removes carrier configuration, as it only held dropoff points


FreekVR (changes requested):
A few questions to double check

NabDevs (approved):

FreekVR (approved):

PR #596, opened 2026-03-06
feat: introduce shipment query, delete, return and multi-collo services

NabDevs:
Migrates remaining MyParcelCollection functionality into dedicated service classes using the generated OpenAPI client as source of truth.

New services:
- ShipmentQueryService (find, findMany, query, findByReferenceId)
- ShipmentDeleteService (deleteMany)
- ReturnShipmentService (createRelated, createUnrelated)
- MultiColloShipmentService (splitShipment)

Refactored services:
- ShipmentTrackTraceService: hybrid approach for street deserialization bug
- ShipmentPrintService: migrated from MyParcelRequest to generated client
- ShipmentLabelsService: removed MyParcelRequest dependency

Services using hybrid approach (request builder + manual send) include @todo markers for reverting once upstream spec/codegen fixes land.

INT-1032


src/Services/CoreApi/ModelHydrator.php
```diff
@@ -0,0 +1,95 @@
...
+            return false;
+        }
+
+        return class_exists($type) && is_subclass_of($type, ModelInterface::class);
+    }
+
+    /**
+     * PHP 7.4-compatible replacement for str_ends_with().
+     */
+    private static function endsWith(string $haystack, string $needle): bool
```
FreekVR: I believe we already have a `Str::endsWith` helper in the SDK? If we don't, it would make more sense to add it to that helper

src/Services/CoreApi/ModelHydrator.php
```diff
@@ -0,0 +1,95 @@
...
+ * pattern in ShipmentDefsShipmentRecipient::setStreet()).
+ *
+ * Without this hydrator, nested objects would remain as plain arrays in the container,
+ * causing failures when consumers call typed getters like getStatus()->getCurrent().
+ *
+ * @todo remove once the upstream spec/codegen fix for the street pattern is available
+ *       and the client is regenerated — at that point ObjectSerializer::deserialize()
+ *       can be used directly without crashing.
+ */
+final class ModelHydrator
```
FreekVR: This whole class is made as a bypass for a specific issue with generated code setters, but then it will also bypass any correct attribute mappings (setter methods) of the generated models? Would it not be better to overwrite the buggy model with a (hopefully temporary) extended model that simply overrides the existing setStreet method OR override the setters() mapping? (Note that you may also want to set `--import-mappings` so that these override models are then used in the `ShipmentApi` class and/or related models - https://openapi-generator.tech/docs/usage#target-external-models)

src/Services/Returns/ReturnShipmentService.php
```diff
@@ -0,0 +1,157 @@
...
+    private PsrClientInterface $httpClient;
+
+    public function __construct(
+        string $apiKey,
+        ?ShipmentApi $api = null,
+        ?PsrClientInterface $httpClient = null,
+        ?string $host = null
+    ) {
+        $this->api = $api ?? ShipmentApiFactory::make($apiKey, $host);
+        $this->httpClient = $httpClient ?? new GuzzleClient(['timeout' => 10]);
```
FreekVR: This timeout number seems a bit random? Suggest we share this between all guzzle clients unless there's a specific reason to have it differently here. (In that case, please document)

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,185 @@
...
+            $parameters['page'] ?? null,
+            $parameters['q'] ?? null,
+            $parameters['reference_identifier'] ?? null,
+            $parameters['region'] ?? null,
+            $parameters['shipment_type'] ?? null,
+            $parameters['shop_id'] ?? null,
+            $size,
+            $parameters['sort'] ?? null,
+            $parameters['status'] ?? null,
+            $parameters['transaction_status'] ?? null
```
FreekVR: Would it not be possible to use a spread or some other mapping here rather than having to define the parameters like this?

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Query shipments by available API filters.
+     *
+     * @param array<string, mixed> $parameters
+     * @return ShipmentDefsShipment[]
+     */
+    public function query(array $parameters = []): array
+    {
+        $size = array_key_exists('size', $parameters) ? $parameters['size'] : 300;
```
FreekVR: Why 300? Please avoid random number:

- Put them in a CONST with a clear name
- Document *why* a certain amount is used

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,185 @@
...
+            $size,
+            $parameters['sort'] ?? null,
+            $parameters['status'] ?? null,
+            $parameters['transaction_status'] ?? null
+        );
+
+        return $this->sendAndParseShipments($request);
+    }
+
+    public function find(int $shipmentId): ?ShipmentDefsShipment
```
FreekVR, line 51: Please add a function description to all service functions, as these will be the main touch points for people using the SDK. (And perhaps in the future, we can generate documentation with it)

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,185 @@
...
+        $result = [];
+
+        foreach ($referenceIdentifiers as $referenceIdentifier) {
+            $referenceIdentifier = trim($referenceIdentifier);
+
+            if ('' === $referenceIdentifier) {
+                continue;
+            }
+
+            $shipments = $this->query(['reference_identifier' => $referenceIdentifier]);
```
FreekVR: Is there no way in the API to do this lookup for multiple records at the same time? This would send an API request for each item in the array (may cause rate limits to be hit)

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,185 @@
...
+     * Recursively hydrate a ShipmentDefsShipment from raw array data.
+     *
+     * Uses ModelHydrator to construct all nested typed models (recipient, sender,
+     * status, options, etc.) via their constructors, bypassing setter validation.
+     */
+    private function hydrateShipment(array $data): ShipmentDefsShipment
+    {
+        /** @var ShipmentDefsShipment */
+        return ModelHydrator::hydrate(ShipmentDefsShipment::class, $data);
+    }
```
FreekVR: Please see my comment elsewhere - suggest overriding the model and/or generated API class instead.

src/Services/TrackTrace/ShipmentTrackTraceService.php
```diff
@@ -4,22 +4,41 @@
...
+    private PsrClientInterface $httpClient;
+
+    public function __construct(
+        string $apiKey,
+        ?ShipmentApi $api = null,
+        ?PsrClientInterface $httpClient = null,
+        ?string $baseUri = null
+    ) {
         $this->api = $api ?? ShipmentApiFactory::make($apiKey, $baseUri);
+        $this->httpClient = $httpClient ?? new GuzzleClient(['timeout' => 10]);
```
FreekVR: See my comment regarding timeout number elsewhere

src/Services/TrackTrace/ShipmentTrackTraceService.php
```diff
@@ -32,35 +51,40 @@ public function fetchTrackTraceData(array $shipmentIds): array
...
+     * Recursively hydrate a ShipmentDefsTrackTrace from raw array data.
+     *
+     * Uses ModelHydrator to construct all nested typed models (recipient, sender,
+     * status, history, location, etc.) via their constructors, bypassing setter validation.
+     */
+    private function hydrateTrackTrace(array $data): ShipmentDefsTrackTrace
+    {
+        /** @var ShipmentDefsTrackTrace */
+        return ModelHydrator::hydrate(ShipmentDefsTrackTrace::class, $data);
+    }
```
FreekVR: See my concern regarding hydration elsewhere

FreekVR (changes requested):

src/Services/CoreApi/ModelHydrator.php
NabDevs: Good catch, replaced with the helper.

src/Services/CoreApi/ModelHydrator.php
NabDevs: Refactored as follows:

- Created `FixedShipmentRecipient extends ShipmentDefsShipmentRecipient` in src/Client/Override/ that overrides only `setStreet()` (corrects the broken regex) and `listInvalidProperties()`.

All marked with @todo remove for when the upstream codegen fix is implemented

src/Services/Returns/ReturnShipmentService.php
NabDevs: Extracted to `ShipmentApiFactory::DEFAULT_HTTP_TIMEOUT = 10 `with documentation explaining the rationale (10s provides headroom for label generation and large batch queries and is also used in Legacy code).

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Moved the parameter mapping into a dedicated `buildGetShipmentsRequest(array $parameters)` method. The query() method itself is now cleaner.

src/Services/CoreApi/ModelHydrator.php
NabDevs: PS: i didn't use the `--import-mapings` option, found it a little overkill as a temp solution for these 2 methods.

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Put it in a CONST with a docblock explaining its origin.

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Added descriptive PHPDoc to all public service methods across all services.

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Documented the N+1 behavior in the PHPDoc. The API only supports filtering by a single `reference_identifier` per request, there's no batch endpoint for this.

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Fixed, see previous hydration override.

src/Services/TrackTrace/ShipmentTrackTraceService.php
NabDevs: fixed, see previous.

src/Services/TrackTrace/ShipmentTrackTraceService.php
NabDevs: Fixed, see previous hydration override.

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Recursively hydrate a ShipmentDefsShipment from raw array data.
+     *
+     * Uses ModelHydrator to construct all nested typed models (recipient, sender,
+     * status, options, etc.) via their constructors, bypassing setter validation.
+     */
+    private function hydrateShipment(array $data): ShipmentDefsShipment
+    {
+        /** @var ShipmentDefsShipment */
+        return ModelHydrator::hydrate(ShipmentDefsShipment::class, $data);
```
FreekVR: See Slack.

I believe you can remove the hydrator by:
- Using the fixed "Override" class directly where appropriate, and
- Registering this class via `--type-mapping` and/or `--import-mappings`
- This should use this class rather than the generated one for any references within the generated clients and models, do please double check the extend still works (the original class should still be created and updated by the generator)

FreekVR (changes requested):

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Fixed, removed the hydrator and used `--type-mapping` to extend the original class with a working one.

FreekVR (approved):

src/Services/MultiCollo/MultiColloShipmentService.php
```diff
@@ -0,0 +1,84 @@
...
+{
+    /**
+     * Convert one shipment into a multi-collo shipment with secondary shipments.
+     *
+     * @param int $amount Total number of collo pieces (including the main shipment).
+     */
+    public function splitShipment(Shipment $shipment, int $amount): Shipment
+    {
+        if ($amount < 2) {
+            throw new InvalidArgumentException('Multi-collo requires at least 2 shipments');
```
joerivanveen: Following offline discussion: we do not have to validate this here.

src/Services/MultiCollo/MultiColloShipmentService.php
```diff
@@ -0,0 +1,84 @@
...
+        if (null !== $weightPerCollo) {
+            $main->setPhysicalProperties(['weight' => $weightPerCollo]);
+        }
+
+        $secondaryShipments = [];
+        for ($i = 1; $i < $amount; $i++) {
+            $secondary = new SecondaryShipmentRequest();
+            $secondary->setReferenceIdentifier($referenceId);
+
+            $carrier = $main->getCarrier();
```
joerivanveen: I would move this assignment to outside the for loop just like weightPerCollo and referenceId. No reason for rejection just an observation.

src/Services/Shipment/ShipmentDeleteService.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Delete one or more shipments by id.
+     *
+     * @param int[] $shipmentIds
+     */
+    public function deleteMany(array $shipmentIds): void
+    {
+        if (empty($shipmentIds)) {
+            throw new InvalidArgumentException('At least one shipment ID is required');
```
joerivanveen: I think you can just return here as well, re: descussion about throwing errors. For consideration.

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Find a single shipment by its reference identifier.
+     *
+     * @return ShipmentDefsShipment|null The shipment, or null if not found.
+     */
+    public function findByReferenceId(string $referenceIdentifier): ?ShipmentDefsShipment
+    {
+        $shipments = $this->findManyByReferenceId([$referenceIdentifier]);
+
+        return $shipments[0] ?? null;
```
joerivanveen: I think findManyByReferenceId may return multiple shipments for the same referenceId, which is a scenario that can happen in the real world I think. Should we not return an array of shipments as well here and let the consuming app decide what to do (error / log / present results) what happens when multiple shipments exist with the same referenceId?

src/Services/Shipment/ShipmentQueryService.php
```diff
@@ -0,0 +1,197 @@
...
+     */
+    private function mergeUniqueById(array $existing, array $incoming): array
+    {
+        $merged = $existing;
+
+        foreach ($incoming as $shipment) {
+            $id = $shipment->getId();
+
+            if (null === $id) {
+                $merged[] = $shipment;
```
joerivanveen: I would consider this a bug, is it possible that a $shipment without an id is returned? If so, we are now silently assigning an id to it here in the $merged array because the keys are id’s, for instance 225 if the previous shipment had id 224, what happens when a shipment with id 225 is also in the loop?

src/Services/TrackTrace/ShipmentTrackTraceService.php
```diff
@@ -11,54 +11,56 @@
...
 
-        foreach ($data->getTracktraces() as $trackTrace) {
-            if (! $trackTrace instanceof ShipmentDefsTrackTrace) {
-                continue;
-            }
+        foreach ($tracktraces as $tracktrace) {
+            $shipmentId = $tracktrace->getShipmentId();
 
-            $result[(int) $trackTrace->getShipmentId()] = $trackTrace;
+            if (null !== $shipmentId) {
```
joerivanveen, line 61: This is a different approach from the ShipmentQueryService, now if there is no shipmentId we just ignore it. Is there a reason why these are different?

test/Services/Shipment/ShipmentPrintServiceTest.php
```diff
@@ -25,99 +31,88 @@ public function testPrintThrowsWhenNoShipmentsWereAdded(): void
...
+                ]));
+            });
 
-        $curlMock->shouldReceive('close')->once();
+        $collection = new ShipmentCollection([
+            (new Shipment())->setReferenceIdentifier('order-1001'),
+        ]);
+
+        $service = new ShipmentPrintService($this->getApiKey(), $api, $httpClient);
+        $service->setUserAgentForProposition('Magento', '2.4.7');
```
joerivanveen: Sanity check: is this useragent necessary for the test, and if so is a more generic one maybe clearer about the intent (for instance 'appname', 'v1.0.0')

joerivanveen (changes requested):

src/Services/MultiCollo/MultiColloShipmentService.php
NabDevs: Removed the validation here, following our discussion.

src/Services/MultiCollo/MultiColloShipmentService.php
NabDevs: Moved the carrier assignment outside the loop, in line with the reference id / weight setup.

src/Services/Shipment/ShipmentDeleteService.php
NabDevs: True, adjusted!

src/Services/Shipment/ShipmentQueryService.php
NabDevs: As discussed offline: I changed findByReferenceId() to return all matches for a single reference identifier.

src/Services/Shipment/ShipmentQueryService.php
NabDevs: Agreed. The merge helper has been removed entirely.

src/Services/TrackTrace/ShipmentTrackTraceService.php
NabDevs: Aligned the query side with this same principle, client = source of truth.

test/Services/Shipment/ShipmentPrintServiceTest.php
NabDevs: Replaced the Magento-specific user agent in the test with a generic value to make the intent clearer.

joerivanveen (approved):

PR #611, opened 2026-04-09
fix: override iam principal deserialization via typeMappings

NabDevs:
Fixes IAM `whoami` deserialization for shop principals by overriding the generated `Principal` parent model.

The generated IAM client currently hardcodes `Principal.role` to `RoleUser`, which causes valid shop responses to fail during deserialization. This PR applies a temporary override via `typeMappings`, similar to the existing fixes in the shipment domain.

This is a temporary workaround for `oneOf` handling in the PHP generator/client. The upstream spec/codegen issue should still be fixed so this override can be removed later.

INT-1494

### DIFF
--- a/openapi/iam.yaml
+++ b/openapi/iam.yaml
@@ -3,6 +3,10 @@
 # OpenAPI Generator configuration
 inputSpec: https://iam.api.myparcel.nl/openapi.json
 outputDir: src/Client/Generated/IamApi
+typeMappings:
+  # @todo remove once upstream Principal generation is fixed
+  # Temporary override for correct whoami Principal deserialization.
+  Principal: FixedPrincipal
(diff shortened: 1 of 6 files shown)


openapi/iam.yaml
```diff
@@ -3,6 +3,10 @@
 # OpenAPI Generator configuration
 inputSpec: https://iam.api.myparcel.nl/openapi.json
 outputDir: src/Client/Generated/IamApi
+typeMappings:
```
FreekVR, line 6: Side note - this issue should be reported to OFF team rather than Alwin / OpenAPI -> they manage their own API specs.

FreekVR (approved):

joerivanveen (approved):

PR #614, opened 2026-04-15
fix: harden v11 shipment service wrappers against spec/client drift

NabDevs:
This PR hardens the v11 shipment service wrappers against recent Core API spec/client changes.

Changes included:
- fix `ShipmentCreateService` to parse the current generated `postShipments` response shape
- fix return shipment creation to parse `data.shipments` instead of the old `data.ids`
- add type mappings for shipment `region` and `payment_status` codegen issues and regenerate the Core API client
- restore label PDF retrieval by using the correct `Accept: application/pdf`
- reduce wrapper fragility by resolving shipment content types by meaning instead of generated array index
- keep existing legacy/direct-print label flows intact where the generated client still does not model the contract correctly

INT-1499

### DIFF
--- a/src/Services/Returns/ReturnShipmentService.php
+++ b/src/Services/Returns/ReturnShipmentService.php
@@ -7,12 +7,15 @@ namespace MyParcelNL\Sdk\Services\Returns;
 use GuzzleHttp\Client as GuzzleClient;
 use InvalidArgumentException;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Api\ShipmentApi;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentDefsShipment;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentPostReturnShipmentsRequest;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentPostReturnShipmentsRequestData;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentPostReturnShipmentsRequestDataReturnShipmentsInner;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentPostUnrelatedReturnShipmentsRequest;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentPostUnrelatedReturnShipmentsRequestData;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentPostUnrelatedReturnShipmentsRequestDataReturnShipmentsInner;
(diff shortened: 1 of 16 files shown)


src/Services/Returns/ReturnShipmentService.php
```diff
@@ -154,4 +189,24 @@ private function withQueryParameter(RequestInterface $request, string $key, stri
         return $request->withUri($uri->withQuery(http_build_query($query)));
     }
 
+    /**
+     * Resolve the generated return shipment content type by prefix instead of array index.
+     *
+     * This keeps the generated client as source of truth while avoiding fragile coupling to
+     * the internal order of ShipmentApi::contentTypes['postShipments'].
+     */
+    private function resolveContentType(string $prefix): string
```
FreekVR: Just for future reference, this could be interesting as a util or helper if needed elsewhere

src/Services/Shipment/ShipmentCreateService.php
```diff
@@ -56,12 +60,26 @@ public function create(ShipmentCollection $collection, $format = null, $position
...
+            $this->resolveShipmentCreateContentType()
         );
 
         return $this->parseCreateResponse($shipments, $response);
     }
 
+    /**
+     * Resolve the generated shipment create content type by prefix instead of array index.
+     */
+    private function resolveShipmentCreateContentType(): string
```
FreekVR: ... and here is the second one already, very similar to the other content type resolver. At this point I'd say this should be reusable :)

src/Services/Shipment/ShipmentPrintService.php
```diff
@@ -59,21 +67,41 @@ public function print(ShipmentCollection $collection, string $printerGroupId): a
...
+            $this->resolveShipmentCreateContentType()
         )->withHeader('Accept', sprintf(self::DIRECT_PRINT_ACCEPT_TEMPLATE, $printerGroupId));
+    }
 
-        $response = $this->httpClient->sendRequest($request);
-        $decoded = json_decode((string) $response->getBody(), true);
+    /**
+     * Resolve the generated shipment create content type by prefix instead of array index.
+     */
+    private function resolveShipmentCreateContentType(): string
```
FreekVR: and this one

FreekVR (changes requested):
Looks good! Just three very similar methods I'd suggest refactoring into a reusable Util or Helper

GravendeelJochem (approved):
Op Freek zn punten na goed!

src/Services/Returns/ReturnShipmentService.php
NabDevs: Good point, I agree. Added a shared helper for this.

FreekVR (approved):

PR #615, opened 2026-04-17
refactor!: remove legacy consignment stack and finalize v11 migration paths

NabDevs:
## Summary

This PR removes the remaining legacy consignment stack from `v11` and completes the migration to generated API clients + typed service wrappers.

See `UPGRADE.md` for the full before/after migration guide.

The goal is to make the v11 SDK fully shipment-first:
- generated clients are the source of truth for API communication
- typed service wrappers replace the old hand-written consignment flows
- legacy code that still has to exist is kept explicit and marked `@internal`
- beta consumers get a clear migration path via `UPGRADE.md` and the updated `README.md`

### Removed
- legacy consignment models
- consignment validators and rules
- consignment encoding / collection / factory helpers
- delivery options adapters and factory
- legacy webhook web services
- `CheckApiKeyWebService`
- consignment-only helpers, exceptions and dead concerns
- orphaned `test/Mock/` directory
- orphaned `LabelHelper`

### Replaced with v11 equivalents
- shipment creation via `ShipmentCreateService`
- shipment query/search via `ShipmentQueryService`
- shipment deletion via `ShipmentDeleteService`
- direct print via `ShipmentPrintService`
- labels via `ShipmentLabelsService`
- track & trace via `ShipmentTrackTraceService`
- returns via `ReturnShipmentService`
- multi-collo via `MultiColloShipmentService`
- webhook management via `WebhookService`
- API key validation via `ApiKeyService`
- fulfilment order shipment options via the shared `ShipmentOptions` model

### Kept for now
The remaining legacy fulfilment/web-service surface is still available where v11 does not yet have a generated-client replacement, and is now explicitly marked `@internal`.

This includes:
- `AccountWebService`
- `CarrierOptionsWebService`
- `PrinterGroupWebService`
- Order v1 fulfilment collections and related legacy support classes
- carrier models and shared legacy infra still required by those flows

## Breaking changes

This PR is intentionally breaking for consumers still using the old consignment API.

Notable removals / migrations:
- `MyParcelCollection` + `AbstractConsignment` -> `ShipmentCollection` + `Shipment`
- `createConcepts()` -> `ShipmentCreateService::create()`
- `find()` / `findMany()` / `findByReferenceId()` / `setLatestData()` -> `ShipmentQueryService`
- `deleteConcepts()` -> `ShipmentDeleteService::deleteMany()`
- legacy webhook classes -> `WebhookService`
- `CheckApiKeyWebService` -> `ApiKeyService`
- delivery options adapters -> ShipmentOptions + Order::setCarrierId()/getCarrierId() for fulfilment Order v1 flows

## Documentation
- added `UPGRADE.md` with before/after migration examples
- updated `README.md` to reflect the v11 service-based API
- clarified which legacy classes still exist and which are removed

INT-1461


GravendeelJochem (approved):

src/Collection/Fulfilment/OrderCollection.php
```diff
@@ -132,45 +133,43 @@ function (Order $order) {
...
-                'amount'   => (int) $shipmentOptions->getInsurance() * 100,
+        if ($options->getInsurance()) {
+            $result['insurance'] = [
+                'amount'   => (int) $options->getInsurance() * 100,
                 'currency' => 'EUR',
             ];
         }
 
-        return $options;
+        return $result;
```
FreekVR, line 180: This implementation would break for any new (future) shipment options, requiring manual updates. Is there a way we can make this more generic - even though the Order v1 api is not in the spec?

src/Helper/SplitStreet.php
```diff
@@ -138,7 +140,7 @@ public static function splitStreet(string $fullStreet, string $local, string $de
      */
     public static function getStreetParts($street)
     {
-        $streetWrap = wordwrap($street, AbstractConsignment::MAX_STREET_LENGTH, 'BREAK_LINE');
+        $streetWrap = wordwrap($street, 40, 'BREAK_LINE');
```
FreekVR: Can we still wrap this in a constant? Prefer no magic numbers

src/Model/Fulfilment/AbstractOrder.php
```diff
@@ -25,9 +24,9 @@ class AbstractOrder extends BaseModel
     public const DATE_FORMAT_DATE = 'Y-m-d';
 
     /**
-     * The selected delivery options for this order.
+     * The selected shipment options for this order.
      *
-     * @var \MyParcelNL\Sdk\Adapter\DeliveryOptions\DeliveryOptionsFromOrderAdapter
+     * @var \MyParcelNL\Sdk\Model\Fulfilment\OrderShipmentOptions
      */
     protected $delivery_options;
```
FreekVR, line 34: Is this correct? Attribute name and new docblock seem like a mismatch. I think we should maybe rename this attribute to avoid confusing.

src/Model/Fulfilment/OrderShipmentOptions.php
```diff
@@ -0,0 +1,320 @@
...
+ *
+ * Replaces the legacy AbstractDeliveryOptionsAdapter + AbstractShipmentOptionsAdapter
+ * hierarchy with a single flat class that covers everything OrderCollection::save() needs.
+ *
+ * Uses the existing ApiMapping classes to translate v2 enum strings (from the generated
+ * client) to v1 integer IDs (required by the Order v1 fulfilment API). When the spec adds
+ * new delivery/package types and the client is regenerated, only the mapping needs a new
+ * entry — this class does not hardcode any IDs or names.
+ */
+class OrderShipmentOptions
```
FreekVR: The `ShipmentOptions` in the fulfilment namespace are actually the same as the `ShipmentOptions` from the shipments namespace. Is there any potential to continue to use the same model/class for both? Or could we continue to use the Shipments namespace to house this model at least?

src/Model/Shipment/Mapping/DeliveryTypeApiMapping.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        return [
+            RefTypesDeliveryTypeV2::MORNING       => RefTypesDeliveryType::MORNING,
+            RefTypesDeliveryTypeV2::STANDARD      => RefTypesDeliveryType::STANDARD,
+            RefTypesDeliveryTypeV2::EVENING       => RefTypesDeliveryType::EVENING,
+            RefTypesDeliveryTypeV2::PICKUP        => RefTypesDeliveryType::PICKUP,
+            RefTypesDeliveryTypeV2::SAME_DAY      => RefTypesDeliveryType::SAME_DAY,
+            RefTypesDeliveryTypeV2::EXPRESS        => RefTypesDeliveryType::EXPRESS,
+            RefTypesDeliveryTypeV2::EARLY_MORNING => RefTypesDeliveryType::EARLY_MORNING,
+        ];
```
FreekVR, line 62: I'd suggest making this map dynamic.

We can match as per this proof of concept code: https://internal.jira.dmp.zone/browse/INT-1441

We can retain a static fallback for any non-matching values, but a dynamic map as default would be good? Perhaps make it reusable via trait or shared abstract class.

src/Services/Labels/ShipmentLabelsService.php
```diff
@@ -191,7 +190,9 @@ private function buildLabelsRequest(array $shipmentIds, string $format, ?string
     private function resolveFormatAndPositions($positions): array
     {
         if (is_numeric($positions)) {
-            return ['A4', LabelHelper::getPositions((int) $positions)];
+            $start = (int) $positions;
+
+            return ['A4', $start >= 1 && $start <= 4 ? implode(';', range($start, 4)) : ''];
```
FreekVR: there are a lot of "magic numbers" here. Suggest making it more readable by adding constants (naming positions) or adding a docblock comment to describe behavior.

FreekVR (changes requested):
A few small requests/enhancements

src/Helper/SplitStreet.php
NabDevs: Agreed, fixed by moving the legacy street length into a local `MAX_STREET_LENGTH` constant.

src/Collection/Fulfilment/OrderCollection.php
NabDevs: Since Order v1 will not be specced, I’d avoid making the payload generic (for now). New generated shipment options are not automatically valid for Order v1, and some fields still need Order v1-specific keys / id conversion.

I agree the explicit mapping could be centralized better, but I’d prefer to keep that as a follow-up, so this PR stays scoped to removing the consignment stack (although it's already scoped beyond that 😋) and preserving current Order v1 behavior.

src/Model/Fulfilment/AbstractOrder.php
NabDevs: I see the mismatch. I’d prefer not to rename the property in this PR because `delivery_options` is part of the legacy Order v1 model shape/hydration. I clarified the docblock to make explicit that the legacy property name now contains `OrderShipmentOptions`.

src/Model/Fulfilment/OrderShipmentOptions.php
NabDevs: This class was added intentionally as a small compatibility bridge for the legacy Order v1 fulfilment flow. It replaces the removed delivery-options adapter hierarchy, so we can delete the consignment stack while keeping `OrderCollection::save()` working. Although it overlaps with `Shipment\ShipmentOptions`, it targets a different contract: Order v1 still expects v1 ids and a fulfilment-specific payload shape. I’d keep it in the fulfilment namespace for now to make that boundary explicit.

src/Model/Shipment/Mapping/DeliveryTypeApiMapping.php
NabDevs: Agreed as a follow-up direction for INT-1441 (I've already assigned the ticket to myself :)). For this cleanup I’d keep the explicit map.

src/Services/Labels/ShipmentLabelsService.php
NabDevs: Agreed, fixed by replacing with named constants. The behavior still matches the old `LabelHelper::getPositions()` behavior.

src/Model/Fulfilment/OrderShipmentOptions.php
FreekVR: Looking at the CoreAPI setup and how this is documented in the developer docs site, the ShipmentOptions for an order are a direct reference to the ShipmentOptions from the (v1) shipments namespace. Unless its a lot of extra work / a completely different PHP contract, I'd push back on this and use the (generated) ShipmentOptions from the Shipments namespace given the *API* contract should be exactly the same as in the API the options definition is shared between shipments and orders.

FreekVR (changes requested):
Just one comment on the earlier thread re: shipment options for `fulfillment` namespace.

src/Model/Fulfilment/OrderShipmentOptions.php
NabDevs: Allright, I changed this to use the shared `MyParcelNL\Sdk\Model\Shipment\ShipmentOptions` wrapper, which extends the generated CoreAPI shipment options contract. The separate `Fulfilment\OrderShipmentOptions` bridge is removed.

Carrier id is now kept separately on the order via `setCarrierId()`/`getCarrierId()`, since carrier is shipment-level data and not part of shipment options. `OrderCollection` still keeps the Order v1 payload mapping explicit, so this does not introduce a broader generic mapper in this PR.

FreekVR (approved):

PR #619, opened 2026-06-08
feat: add carrier contract definitions service

NabDevs:
Adds `CarrierContractDefinitionsService` as a thin public wrapper around the generated Core API contract definitions endpoint.

This gives integrations a v11-safe way to fetch account-level carrier contract definitions without using the legacy `CarrierOptionsWebService` / `CarrierFactory` path.

## Changes

- Added `CarrierContractDefinitionsService::getAll()`
- Added `CarrierContractDefinitionsService::getByCarrier()`
- Returns generated Core API contract definition models directly
- Added unit coverage with mocked `ShipmentApi`

### DIFF
--- a/src/Services/Capabilities/CarrierContractDefinitionsService.php
+++ b/src/Services/Capabilities/CarrierContractDefinitionsService.php
@@ -25,25 +25,40 @@ final class CarrierContractDefinitionsService
 
     private ShipmentApi $api;
 
+    /**
+     * @param string           $apiKey API key used when creating the default generated client.
+     * @param string|null      $host   Optional API host override for the default generated client.
+     * @param ShipmentApi|null $api    Optional preconfigured generated client, primarily for tests or custom transports.
+     *                                 When provided, the client is used as-is.
+     */
     public function __construct(
         string $apiKey,
-        ?ShipmentApi $api = null,
-        ?string $host = null
+        ?string $host = null,
+        ?ShipmentApi $api = null
     ) {
-        $this->api = $api ?? ShipmentApiFactory::make($apiKey, $host);
+        $this->api = $api ?? ShipmentApiFactory::make($apiKey, $host, $this->getUserAgentHeader());
     }
 
     /**
-     * Fetch all contract definitions available for the configured API key.
+     * Fetch contract definitions available for the configured API key.
+     *
+     * Pass a generated carrier enum value to let the generated request model
+     * validate and filter the API request.
      *
      * @return RefCapabilitiesContractDefinitionsResponseContractDefinitionsV2[]
      * @throws \MyParcelNL\Sdk\Client\Generated\CoreApi\ApiException
      */
-    public function getAll(): array
+    public function getContractDefinitions(?string $carrier = null): array
     {
+        $request = new CapabilitiesPostContractDefinitionsRequestV2();
+
+        if (null !== $carrier) {
+            $request->setCarrier($carrier);
+        }
+
         $response = $this->api->postCapabilitiesContractDefinitions(
             $this->getUserAgentHeader(),
-            new CapabilitiesPostContractDefinitionsRequestV2()
+            $request
         );
 
         if (! $response instanceof CapabilitiesResponsesContractDefinitionsV2) {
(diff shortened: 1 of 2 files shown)


src/Services/Capabilities/CarrierContractDefinitionsService.php
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public function getByCarrier(string $carrier): array
+    {
+        $normalizedCarrier = $this->normalizeCarrier($carrier);
+
+        return array_values(array_filter(
+            $this->getAll(),
+            fn (RefCapabilitiesContractDefinitionsResponseContractDefinitionsV2 $definition): bool => $normalizedCarrier ===
+                $this->normalizeCarrier((string) $definition->getCarrier())
+        ));
```
FreekVR: If I'm reading this right this filters the full results, instead of just passing the carrier filter into the request, which would return only results for that carrier and not require in-memory filtering (faster). 

The docblock explanation for this makes little sense, given the SDK (generated client) itself governs which carriers are/are not supported. I'd keep the methods here as a simple wrapper/helper around the different types of request rather than doing local filtering.

In fact, an optional $carrier parameter for `getAll` (perhaps rename to `getContractDefinitions`?) would suffice rather than having to maintain two distinct methods for a simple optional API param.

src/Services/Capabilities/CarrierContractDefinitionsService.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private function normalizeCarrier(string $carrier): string
+    {
+        return strtoupper((string) preg_replace('/[^a-z0-9]/i', '', $carrier));
+    }
```
FreekVR: I feel like it would be better not to use this method to ensure people pass the correct carrier ID rather than (attempting to) correct their mistakes for them.

src/Services/Capabilities/CarrierContractDefinitionsService.php
```diff
@@ -0,0 +1,83 @@
...
+    use HasUserAgent;
+
+    private ShipmentApi $api;
+
+    public function __construct(
+        string $apiKey,
+        ?ShipmentApi $api = null,
+        ?string $host = null
+    ) {
+        $this->api = $api ?? ShipmentApiFactory::make($apiKey, $host);
```
FreekVR: if a `ShipmentApi $api` parameter is passed, `$apiKey` and `$host` effectively become dead.

I'd suggest:
1. Swapping `$api` and `$host` parameters so that the ordering makes more sense (no awkward `'foo.bar', null, 'myapikey'`-arguments needed if you want to pass a key and host but NOT a prebuilt client)
OR
2. Setting the `$apiKey` and `$host` on the existing `ShipmentApi $api` within the consturctor

In either case, clearly document the behavior in the constructor docblock.

src/Services/Capabilities/CarrierContractDefinitionsService.php
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Fetch all contract definitions available for the configured API key.
+     *
+     * @return RefCapabilitiesContractDefinitionsResponseContractDefinitionsV2[]
+     * @throws \MyParcelNL\Sdk\Client\Generated\CoreApi\ApiException
+     */
+    public function getAll(): array
+    {
+        $response = $this->api->postCapabilitiesContractDefinitions(
+            $this->getUserAgentHeader(),
```
FreekVR, line 69: It might be possible to pass a default UA string into the `ShipmentApiFactory` for consistency?

FreekVR (changes requested):

GravendeelJochem (approved):

src/Services/Capabilities/CarrierContractDefinitionsService.php
NabDevs: Agree, removed this.

src/Services/Capabilities/CarrierContractDefinitionsService.php
NabDevs: Fair, adjusted this to remove the local filtering methods and kept it thin. Client = source of truth for validaiton.

src/Services/Capabilities/CarrierContractDefinitionsService.php
NabDevs: Okay, other services currently use this same order, but the constructor is now `apiKey, host, api`, and the PHPDoc clarifies that apiKey/host are only used when the service creates the default client. If a preconfigured ShipmentApi is passed, it is used as-is.

src/Services/Capabilities/CarrierContractDefinitionsService.php
NabDevs: I pass the default `getUserAgentHeader()` into `ShipmentApiFactory::make()` when the service creates its own generated client. Proposition-specific user agents set later are still passed per request through the endpoint’s required `user_agent` argument.

FreekVR (approved):

PR #639, opened 2026-09-02
feat(connect): add connect config, state and storage contract

FreekVR:
Adds the configuration, the stored state and the encryption MyParcel Connect needs. Nothing calls them yet; `ConnectService` follows in the next PR of the stack. Stacked on #643.

Added `ConnectException` with one named constructor per failure the flow can produce, so a consumer catches one type. `fromCallbackError()` maps the four callback values and treats anything else as a server error, because the callback relays whatever the identity provider produced.

Added `Crypto/AesGcmCipher`, which encrypts the private key and the access token before they reach consumer storage. The envelope is `version | iv | tag | ciphertext`, with the version carried as additional authenticated data so it cannot be swapped for another one unnoticed.

Added `Model/Connect/ConnectConfig`, the only place in the SDK that builds an e-commerce host. It takes a platform and derives the host label from it, so there is one value to get right instead of two.

Added `ConnectPlatform`, which re-exports the platform values from the generated `ConnectStartConfig` and maps each one to its host label. The label is not the lowercased platform name: `WOOCOMMERCE` is `woo`. `CSCART` and `LIGHTSPEED` are left out until a PHP plugin targets one, so a platform without a service fails in the constructor instead of at the first call.

Added the three stored records, split by lifetime: `ConnectInstallation` holds the key pair, the connection id and the token challenge for as long as the shop exists; `ConnectToken` holds nothing that outlives it, so a cache with a lifetime is safe; `ConnectNonces` is short lived and safe to lose.

Added `Services/Connect/ConnectStorageInterface`, which the consumer implements. Each save method says where its record belongs and how long to keep it. The two secret-bearing getters say so in their names, so an implementation can see it is storing ciphertext.

Added `Services/Connect/ConnectStateRepository`, the only class that knows both sides. It writes the two nonces per field rather than per record, because a background refresh and a connect flow can be in flight at once and a whole-record write would drop the other's field.

Added `ConnectTokenResponse` over the generated 200 models, with `fromExchange()` and `fromRefresh()`, because a refresh answer does not repeat the connectionId. It carries the two things the generated models cannot: the `DPoP-Nonce` response header, and the `expiresIn` lower bound that the spec declares but the generator does not emit.

`ConnectScope`, `ConnectCallbackError` and `ConnectErrorCode` stay hand written. The spec types scope as a plain string, and the callback goes to the shop rather than to MyParcel, so it describes neither.

Refs INT-1837


src/Model/Connect/ConnectErrorCode.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Sdk\Model\Connect;
+
+/**
```
FreekVR, line 14: Is there any overlap / reuse potential here with the generated openapi client code?

src/Model/Connect/ConnectNonces.php
```diff
@@ -0,0 +1,125 @@
...
+    private $dpopNonce;
+
+    public function __construct(?string $startNonce = null, ?int $startNonceExpiresAt = null, ?string $dpopNonce = null)
+    {
+        $this->startNonce          = $startNonce;
+        $this->startNonceExpiresAt = $startNonceExpiresAt;
+        $this->dpopNonce           = $dpopNonce;
+    }
+
+    public function getStartNonce(): ?string
```
FreekVR, line 61: Not all public methods here have usage docs

src/Crypto/AesGcmCipher.php
```diff
@@ -0,0 +1,138 @@
...
+            OPENSSL_RAW_DATA,
+            $iv,
+            $tag,
+            $version,
+            self::TAG_LENGTH
+        );
+
+        if (false === $ciphertext) {
+            throw ConnectException::invalidArgument('Could not encrypt the connect state');
+        }
```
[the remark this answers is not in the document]
FreekVR: Correct on both counts. Every input to that call is fixed, validated, or freshly generated: the cipher is a constant that `isSupported()` already confirms, the key is checked in the constructor, the IV is 12 random bytes, and the plaintext is arbitrary by design. A `false` there cannot be a caller's argument.

It was also inconsistent with the other half of the class: `decrypt()` throws `decryptionFailed()` on all three of its paths. Added `ConnectException::encryptionFailed()` and used it, so the pair reads the same way round.

No test for it, and there cannot be one without a seam in production code that exists only for the test. It was untested before for the same reason.

src/Model/Connect/ConnectInstallation.php
```diff
@@ -0,0 +1,175 @@
...
+            return null;
+        }
+
+        return new self(
+            (string) $data['encryptedPrivateKey'],
+            self::optionalString($data, 'connectionId'),
+            self::optionalString($data, 'challengeHtm'),
+            self::optionalString($data, 'challengeHtu')
+        );
+    }
```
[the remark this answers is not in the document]
FreekVR: Correct, and worse than the comment says. Reproduced all four cases: `''` throws, `0` gives a key of `'0'`, `true` gives `'1'`, and an array gives `'Array'` plus a PHP warning. Only an absent field read as absent.

The three cast cases are the worse half, because they hand back an installation whose key is nonsense and the failure then surfaces much later as a decryption error blaming the encryption key.

`fromArray()` now reads a non-string or empty envelope as absent, matching its own documented contract, with the cast dropped. The constructor keeps its guard for callers who build one directly. Five tests, written failing first, cover the empty case and a non-string integer, boolean, array and float.

Worth noting the call site is a consumer's: `ConnectStorageInterface`'s docblock tells implementers to write exactly `ConnectInstallation::fromArray(json_decode($raw, true))`, so a truncated stored value came straight in here, and `isConnected()` promises never to throw.

src/Model/Connect/ConnectToken.php
```diff
@@ -0,0 +1,132 @@
...
+                return null;
+            }
+        }
+
+        return new self(
+            (string) $data['encryptedAccessToken'],
+            (int) $data['expiresAt'],
+            (string) $data['scope']
+        );
+    }
```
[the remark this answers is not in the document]
FreekVR: Correct. Reproduced: an empty `encryptedAccessToken` throws, an array token or scope becomes `'Array'`, a non-numeric `expiresAt` becomes `0`, which reads as expired at the epoch, and an array expiry becomes `1`.

`fromArray()` now reads any of those as absent. One case had to keep working: a numeric string `expiresAt`. `HasRecordVersion` already documents why, since a per-column table or a cache layer hands back `'1'` instead of `1`, so the check is `is_numeric` rather than a type test. Six tests written failing first, plus one pinning the stringified expiry.

src/Services/Connect/ConnectStateRepository.php
```diff
@@ -0,0 +1,182 @@
...
+     * nonce of a connect flow the merchant is still in, and the callback would be refused.
+     */
+    public function saveStartNonce(ConnectState $state): void
+    {
+        $nonces = $this->storage->loadNonces() ?? new ConnectNonces();
+
+        $this->storage->saveNonces(
+            (new ConnectNonces($state->getStartNonce(), $state->getStartNonceExpiresAt(), $nonces->getDpopNonce()))
+        );
+    }
```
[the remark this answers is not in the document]
FreekVR: Left as it is. The one production caller, `ConnectService::start()`, builds the nonce from `random_bytes()` immediately before this, so a null cannot reach here, and `burnStartNonce()` already exists for the clearing case. A null guard would turn a programming error into a silent no-op at the moment it matters most, and the repo rule is to add a check only where the input that needs it has actually been seen.

If it should be impossible rather than tolerated, the fix is the signature, `saveStartNonce(string $startNonce, int $expiresAt)`, since the method only reads those two fields off the state. Happy to do that; it touches the caller in the pull request above this one.

src/Exception/ConnectException.php
```diff
@@ -0,0 +1,229 @@
...
+        return self::withCode('session_expired', 'The connect session expired, start again');
+    }
+
+    /**
+     * MyParcel could not match the shop to a sales channel.
+     */
+    public static function invalidSalesChannel(): self
+    {
+        return self::withCode('invalid_sales_channel', 'MyParcel could not link this shop to a sales channel');
+    }
```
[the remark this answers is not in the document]
FreekVR: Correct, and the drift would be silent rather than loud, which is what makes it worth fixing. `ConnectService::translate()` switches on `getErrorCode()` against `ConnectErrorCode::*` and `nonceFrom()` compares against `ConnectErrorCode::USE_DPOP_NONCE`, so a literal here diverging from a constant there would stop the nonce retry firing and stop `invalid_grant` being read per endpoint, with no error anywhere.

All six now reference constants. Two of them, `invalid_sales_channel` and `server_error`, exist on both classes, so I resolved it by where the value is consumed: exception codes are compared through `ConnectErrorCode`, so that one wins wherever it has the value, and `ConnectCallbackError` covers `access_denied` and `session_expired`, which only ever arrive on the callback.

Added the mirror of the existing callback round-trip test: every code `ConnectException` produces has to be a value declared across the two classes, so a seventh named constructor with an undeclared literal fails. Verified it catches one by typoing `server_error`.

src/Model/Connect/ConnectErrorCode.php
FreekVR: Total overlap, and it is now sourced from the generated code. All six values are declared there, but spread across six response models with one value each, so there was nothing to import wholesale. Each constant now points at the model that declares it, which makes this the one place to read them from.

I had an objection to that and it was wrong, so worth recording: I thought referencing the numbered `ConnectTokenPost400ResponseAnyOf1` and friends would silently repoint a constant if the spec reordered its `anyOf` branches. It would not. The constant name encodes the value, so `AnyOf1` carries only `ERROR_INVALID_GRANT`, and a reorder makes that reference stop existing rather than change value. Verified: it raises an `Error` at class load.

What is not usable is the aggregate `getErrorAllowableValues()` on those models. The generator collapses each `anyOf` and keeps only the last branch, so the token 400 reports one allowable error out of four and the refresh 400 one out of two. The individual constants are correct, which is why this references those and not the method. Left alone deliberately: nothing reads those two models, since `asConnectException()` json_decodes the raw body, so a `Fixed*` class for each would be surface with no consumer. Added to what has gone to the team that owns the spec, alongside the `oneOf` collapse on the 202.

The docblock also promised replacement by a generated model once the spec was published. That has happened, so it now says what this is instead.

The existing test asserting the six literal strings has become a drift detector rather than a tautology: the constants come from generated code now, so a regeneration that changes a value fails it. Verified by changing one.

src/Model/Connect/ConnectNonces.php
FreekVR: Right: five of the eight had none. The pattern was that I wrote one wherever there was something non-obvious to say and left the rest bare, which meant a reader had to work out from the class docblock that the two getters return different kinds of thing, one ours and one MyParcel's, rather than from the method in front of them.

Done for all three stored records rather than this one, since `ConnectInstallation` and `ConnectToken` had the same gaps and fixing one of three would have left the set inconsistent. Twenty four public methods, each leading with a verb, in the same format as the `ConnectService` methods.

That also reworded the descriptions that existed but read as noun phrases, so the three files are consistent: "The key pair, encrypted" became "Get the encrypted key pair", and `withTokenChallenge`'s paragraph now opens with what it does before explaining why the challenge lives on the installation rather than the token.

NabDevs (approved):

src/Crypto/AesGcmCipher.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    public function __construct(string $key, ?callable $randomBytes = null)
+    {
+        self::assertOpensslIsLoaded();
+
+        if ('' === $key) {
+            throw ConnectException::invalidArgument('The encryption key cannot be empty');
+        }
+
+        $this->key         = hash('sha256', $key, true);
```
joerivanveen, line 69: I am not saying that I fully understand this, but should 'sha256' maybe be a constant, since it ‘depends’ maybe on something, like 'aes-256-gcm'? Not blocking for me.

src/Model/Connect/ConnectConfig.php
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * What the consumer sets once, before any connect call.
+ *
+ * Immutable: every with* method returns a copy, so a config can be shared without anything changing
+ * it underneath.
+ */
+final class ConnectConfig
+{
+    /**
+     * The only two host shapes. This class is the only place in the SDK that builds an e-commerce
```
joerivanveen, line 18: This seems buried now, `Model/Connect` does not mention the ecommerce api. I suppose we are not using `MyParcelRequest.php` anymore, where the core api and printing endpoint were a constant (also not ideal). Should endpoint usage / availability and the switching between production and acceptance be centrally managed / set in the sdk? Maybe a separate ticket, not for this story obviously.

src/Model/Connect/ConnectToken.php
```diff
@@ -0,0 +1,149 @@
...
+    public static function fromArray(?array $data): ?self
+    {
+        $encryptedAccessToken = self::optionalString($data ?? [], 'encryptedAccessToken');
+        $scope                = self::optionalString($data ?? [], 'scope');
+        $expiresAt            = $data['expiresAt'] ?? null;
+
+        // Nothing usable reads as absent rather than throwing or carrying a cast value: an array
+        // becomes 'Array', and a non-numeric expiry becomes 0, which reads as expired at the epoch.
+        // A numeric string is accepted for the same reason hasKnownVersion() accepts one.
+        if (!self::hasKnownVersion($data)
```
joerivanveen, line 130: It looks like here the original `$data` is silently upgraded to the current version (`new self` below...), I guess this is because of the 'upgrade path' mentioned in `hasKnownVersion` and as such intentional. Just checking.

src/Services/Connect/ConnectStateRepository.php
```diff
@@ -0,0 +1,254 @@
...
+        if ($fresh || !$this->hasLoaded) {
+            $this->loaded    = $this->read();
+            $this->hasLoaded = true;
+        }
+
+        return $this->loaded;
+    }
+
+    /**
+     * The same, for a caller that cannot work without one.
```
joerivanveen: 'the same' does not look great for hints in IDE's, but I guess nobody works there anymore :-P

test/Services/Connect/ConnectStateRepositoryTest.php
```diff
@@ -0,0 +1,364 @@
...
+
+    public function testAReadIsRememberedSoSigningOneCallDoesNotReadTwice(): void
+    {
+        // Signing an API call reads the state twice, and an order import signs hundreds.
+        $this->repository->saveInstallation($this->fullState());
+        $before = $this->reads();
+
+        $this->repository->load();
+        $this->repository->load();
+        $this->repository->load();
```
joerivanveen, line 48: the coincidence that you load three times and the fact that there are 3 storage keys being read each load(), made me question the validity of the test. I would add one more `$this->repository->load();` to make explicit that the number of loads has nothing to do with the literal 3 (and is in fact proving that it works).

joerivanveen (approved):

src/Model/Connect/ConnectConfig.php
FreekVR: This specific class isn't really buried, it's mentioned in docs and type hinted where you need it. It has a documented. `withAcceptance()` method that will facilitate acceptance hosts.

I agree a generic prod/acc switch in the future could be useful, we'd have to introduce a global SDK config and have other classes listen for it. This implementation does not make that impossible and leaves that option available.

src/Crypto/AesGcmCipher.php
FreekVR: This class itself will only ever produce an AES cypher, and moving it to a const would imply that other encryption methods would work, where they would not.

src/Services/Connect/ConnectStateRepository.php
FreekVR: Agreed, I'll revise this.

src/Model/Connect/ConnectToken.php
FreekVR: It's effectively dead code anyway, since there is only one version now.

That's why `hasKnownVersion` holds a comment that a version bump would require an upgrade branch in this very method, and because there is no upgrade, it holds none now.

test/Services/Connect/ConnectStateRepositoryTest.php
FreekVR: I'll add a fourth, since there's no harm.

PR #641, opened 2026-09-02
feat(connect): add the connect service and its transport

FreekVR:
Adds the MyParcel Connect flow: starting it, handling the callback, and keeping a usable access token. `DpopMiddleware` and the docs follow in the rest of the stack. Stacked on #639.

Added `ConnectService` with `start()`, `handleCallback()`, `getAccessToken()`, `refresh()`, `isConnected()`, `disconnect()` and `uninstall()`. `getAccessToken()` refreshes on its own, so no caller has to think about expiry, and `isSupported()` reports whether this PHP installation can run Connect at all.

The two connect calls go through the generated `EcommerceApi` client, so the paths, the request bodies, the response models and the `/connect/start` query all come from the spec. The service adds only what the generated code does not do: it reads the `DPoP-Nonce` response header, which appears nowhere in the generated tree, and converts the generated `ApiException` into a `ConnectException` so a consumer catches one type.

The third constructor argument is a Guzzle client, so a consumer sets a proxy, a CA bundle or a timeout directly. Connect sends the same `User-Agent` as the rest of the SDK through `Concerns\HasUserAgent`, including a plugin registered with `setUserAgentForProposition()`.

The nonce retry runs once per call and keeps what MyParcel sends, so only the first call pays for it. A repeated demand is reported as a fault rather than retried again.

`invalid_grant` is read per endpoint: at the exchange the code is gone and any existing token stays, at the refresh the connection is gone and the token is cleared.

The callback's `htm` and `htu` are checked before they are stored. An `htu` carrying a query cannot be signed, and storing one would make every later refresh fail with no way back but reconnecting.

`disconnect()` also clears the token challenge, so a scheduled refresh cannot reconnect a shop the merchant just disconnected.

`start()` trims the shop name rather than refusing it. The name is the platform's own shop title, which a plugin does not choose, so a long one is cut to 64 characters and a control character dropped. The shop URL is refused instead of altered, because the callback path is appended to it. The shop name's own bounds are checked here rather than with the generated model, which emits the spec's pattern without preg's `/u` flag and so refuses `Straße` and `Ølsalg` while passing `Müller`.

Declared `psr/http-client` and `psr/http-message`, which three existing services already import.

Refs INT-1837


src/Services/Connect/ConnectService.php
```diff
@@ -0,0 +1,844 @@
...
+
+        try {
+            // Read only. Verifying the signature is the resource server's job, and it needs keys we
+            // do not have.
+            $claims = json_decode(Str::base64UrlDecode($segments[1]), true);
+        } catch (Throwable $exception) {
+            throw ConnectException::keyMismatch();
+        }
+
+        $boundTo = is_array($claims) ? ($claims['cnf']['jkt'] ?? null) : null;
```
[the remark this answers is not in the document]
FreekVR: Already handled, and by the operator the comment overlooks. The line is `is_array($claims) ? ($claims['cnf']['jkt'] ?? null) : null`, and `??` uses isset semantics, so it short-circuits the whole chain rather than just the last key. A string, an int, a bool, a null and a list all yield `null` with no warning, and `!is_string($boundTo)` then raises `keyMismatch()`.

Checked against every shape a `cnf` claim could hold on PHP 7.4, 8.0 and 8.4, with an error handler installed to catch any notice. None fired on any version.

No change.

src/Exception/ConnectException.php
```diff
@@ -187,6 +188,35 @@ public static function httpError(int $statusCode, array $headers = [], ?array $b
...
+    /**
+     * Copy the HTTP detail of a failed call onto a case that names it better.
+     *
+     * So a translated failure still reports the status, the body and the value MyParcel actually
+     * wrote, which is what support needs, while the message says what it means.
+     */
+    public function withHttpContextOf(self $source): self
+    {
+        // A new one rather than a clone: PHP exceptions are uncloneable.
+        $exception                  = new self($this->getMessage(), $this->getCode());
```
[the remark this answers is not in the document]
FreekVR: Right, and there were two broken links rather than one.

The one you name: `withHttpContextOf()` built the new exception with two constructor arguments and never passed the source as the third, so `getPrevious()` was null.

The one a step earlier, and the more useful of the two: `asConnectException()` read the headers, code and body off the `ApiException` and then discarded it. That exception is where the diagnostic message lives, since the generated client builds it as `[code] <Guzzle's message>`, carrying the method, the full URL and part of the response body. None of our own messages carry any of that, and it was only being kept on the network-failure branch.

Both now chain, and `httpError()` and `networkFailure()` take an optional previous. A revoked connection on a refresh reads:

```
ConnectException: The connection was revoked, the shop has to connect again
  ConnectException: MyParcel answered 400
    ApiException: [400] Client error: `POST https://generic.ecommerce.api.acceptance.myparcel.nl/connect/refresh` resulted in a ...
```

with `getStatusCode()`, `getErrorCode()` and `getResponseBody()` unchanged on the outermost one.

src/Crypto/AesGcmCipher.php
```diff
@@ -43,30 +43,23 @@ final class AesGcmCipher
...
      *                                                    ConnectException, so a consumer that starts
      *                                                    there never sees this one.
      */
-    public function __construct(string $key, ?callable $randomBytes = null)
+    public function __construct(string $key)
     {
         self::assertOpensslIsLoaded();
 
         if ('' === $key) {
             throw ConnectException::invalidArgument('The encryption key cannot be empty');
```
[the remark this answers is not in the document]
FreekVR: Does not apply, we are not changing any already-released contract here: `AesGcmCipher` is added by this stack and does not exist on `v11`.

src/Crypto/DpopProofFactory.php
```diff
@@ -40,28 +40,12 @@ final class DpopProofFactory
...
-        ?callable $clock = null,
-        ?callable $randomBytes = null
-    ) {
-        $this->signer      = $signer ?? new OpensslJwsSigner();
-        $this->clock       = $clock ?? 'time';
-        $this->randomBytes = $randomBytes ?? 'random_bytes';
+    public function __construct(?JwsSignerInterface $signer = null)
+    {
+        $this->signer = $signer ?? new OpensslJwsSigner();
     }
```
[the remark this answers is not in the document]
FreekVR: Does not apply, same as the `AesGcmCipher` comment: `DpopProofFactory` is added at the bottom of this stack and does not exist on `v11`, so there is no released contract to break. Both parameters were removed on purpose, because they existed only to make tests deterministic.

src/Services/Connect/ConnectService.php
```diff
@@ -0,0 +1,830 @@
...
+    /**
+     * Get an access token that is valid now. Refreshes the stored one when it is near expiry.
+     *
+     * @throws \MyParcelNL\Sdk\Exception\ConnectException When the shop is not connected.
+     */
+    public function getAccessToken(): string
+    {
+        $state = $this->repository->loadOrFail();
+
+        if (null === $state->getAccessToken() || null === $state->getTokenChallenge()) {
```
NabDevs, line 296: The access tokens are cached, but what if a cache is emptied? Shouldn't it do a refresh as well then?

NabDevs (changes requested):

src/Services/Connect/ConnectService.php
FreekVR: The storage is (meant to be) permanent, and a cache is not a secure enough storage option for access tokens.
The `ConnectStorageInterface` further documents how long/permanent things need to be stored.

I checked the touch points in the storage interface that could cause an access token to be empty:

-`saveToken` requires a valid `ConnectToken` or null. That class has several safeguards making sure the access token is a non-empty string. The `null` save-state is meant for disconnecting the shop, and the interface documents this.
-`loadToken` handles the loading side of this.
-`clear` is meant for a full uninstall, also documented, and clears the access token along with everything that encrypts it.

External factors *could* clear the storage, but in that case I'd argue we should probably error out anyway since we can't verify what happened.

NabDevs (approved):

src/Services/Connect/ConnectService.php
```diff
@@ -0,0 +1,830 @@
...
+    /**
+     * The nonce MyParcel sent, if it sent one.
+     *
+     * @param array<string, string[]> $headers
+     */
+    private static function nonceIn(array $headers): ?string
+    {
+        // The first value, never getHeaderLine()'s ", " joined list: that string is refused as a
+        // nonce every time, which would put the flow into a permanent retry loop.
+        $value = $headers[self::NONCE_HEADER][0] ?? null;
```
joerivanveen, line 461: Are we certain CoPilot did not have a good point? Maybe reverse proxies and network infrastructure could change the key `DPoP-Nonce` to lowercase and that would break here?

src/Services/Connect/ConnectService.php
```diff
@@ -0,0 +1,830 @@
...
+            } catch (ConnectException $retryFailure) {
+                if (null !== self::nonceFrom($retryFailure)) {
+                    throw ConnectException::nonceRequired();
+                }
+
+                throw $this->translate($retryFailure, $endpoint);
+            }
+        }
+
+        // MyParcel can rotate the nonce on a successful answer too, to save the next call a trip.
```
joerivanveen, line 512: Nice!

test/Services/Connect/ConnectServiceStartTest.php
```diff
@@ -0,0 +1,510 @@
...
+        try {
+            $this->service()->start('My Shop', 'http://shop.example.test');
+            self::fail('expected a ConnectException');
+        } catch (ConnectException $exception) {
+            self::assertNull($this->storage->row('nonces'));
+            self::assertNull($this->storage->row('installation'));
+        }
+    }
+
+    public function testStartRefusesDetailsThatAreTooLongTogether(): void
```
joerivanveen, line 111: I saw this in earlier prs as well, do we really want to block this, is it not possible to truncate here?

joerivanveen (changes requested):

test/Services/Connect/ConnectServiceStartTest.php
FreekVR: I'd rather error out, since these details are shown on a a consent screen for security reasons.

Truncating would allow someone to just supply a huge string that would then effectively hide the URL:

Eg.
`https://www.averytrustworthywebsiteTHATACTUALLYSTEALSYOURCREDENTIALS.com`

would show as

`https://www.averytrusthworthywebsite...`

That's not to say that the SDK user could just bypass that, but from the SDK side I'd rather not offer that security nit.

src/Services/Connect/ConnectService.php
FreekVR: Yes, I've been fighting AI over this.

We currently have nothing in the SDK or PDK that converts headers to be case-insensitive, neither outgoing nor incoming (though the latter is more limited).

I felt it would lead to overengineering now, but if enough humans agree with the AIs here, I'll pick it up.

joerivanveen: As discussed, we leave it for now.

test/Services/Connect/ConnectServiceStartTest.php
joerivanveen: Ok agreed.

joerivanveen (approved):

PR #642, opened 2026-09-07
feat: derive API mappings from generated definitions

NabDevs:
The SDK v11 now uses fixed tables to convert carrier, delivery type, and package type values. These tables require manual changes when the API adds values.

This change adds `ApiMapperService`. The service uses shared constant names from generated API definitions to convert between:
- Delivery options names.
- API v1 IDs.
- API v2 names.

Explicit mapping rules take priority. If no rule applies, the service matches constant names. For conversions between names, it can also change the letter case and separators. The result must exist in the target definition. If no valid result exists, the service returns null.

The existing mapping classes use the new service. They keep their exceptions for unknown values and preserve the order of existing entries.

The capabilities request mapper also uses generated property definitions. Explicit rules handle renamed options. This change restores three missing option mappings and preserves accepted input formats. The tracked option remains unsupported by the capabilities v2 request.

Separate changes are required for PDK, OpenCart, Shopware, and Magento. 

Fixes INT-1441


src/Model/Capabilities/CapabilitiesMapper.php
```diff
@@ -159,20 +159,36 @@ public function mapFromCoreApi(CoreResponseV2 $core): CapabilitiesResponse
...
+     * @param  array<string, mixed> $optionsData
+     *
+     * @return \MyParcelNL\Sdk\Client\Generated\CoreApi\Model\CapabilitiesOptionsV2
      */
     private function mapOptions(array $optionsData): CoreOptionsV2
     {
-        $options = new CoreOptionsV2();
+        $options          = new CoreOptionsV2();
+        $setters          = CoreOptionsV2::setters();
+        $wireToLocalNames = array_flip(CoreOptionsV2::attributeMap());
```
FreekVR: Claude really likes to name things "wire"

Both `wire` and `local` are unclear, so I can't really understand what this maps from and to here.

Name maps to what they map to and from, I *suspect* this would be a `v2ToDeliveryOptions`-map?

src/Model/Capabilities/CapabilitiesMapper.php
```diff
@@ -159,20 +159,36 @@ public function mapFromCoreApi(CoreResponseV2 $core): CapabilitiesResponse
...
+        $setterNames      = array_combine(array_map('strtolower', $setters), $setters);
 
         foreach ($optionsData as $key => $value) {
-            $setter = self::KNOWN_OPTION_SETTERS[$key] ?? $this->getFallbackSetterName($key);
+            if (in_array($key, self::UNMAPPABLE_OPTIONS, true)) {
```
FreekVR: Are unsupported/unmappable options suffuciently covered in callers of this method? E.g. whomever/whatever uses this, needs to (be able to) decide what to do in this case, because they're silently dropped here.

src/Model/Capabilities/CapabilitiesMapper.php
```diff
@@ -159,20 +159,36 @@ public function mapFromCoreApi(CoreResponseV2 $core): CapabilitiesResponse
...
+                continue;
+            }
+
+            // Accept generated local names and public wire names.
+            $targetProperty = self::KNOWN_OPTION_ALIASES[$key] ?? $wireToLocalNames[$key] ?? $key;
+
```
FreekVR, line 178: Again, "local name" and "wire" terminology are a bit confusing here.

src/Model/Shipment/Mapping/CarrierApiMapping.php
```diff
@@ -5,67 +5,45 @@
 namespace MyParcelNL\Sdk\Model\Shipment\Mapping;
 
 use InvalidArgumentException;
-use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefTypesCarrier;
-use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefTypesCarrierV2;
-
+use MyParcelNL\Sdk\Services\Mapping\ApiMapperService;
+
+/**
+ * Preserves the existing v2-to-id API and its exceptions for unknown values.
```
FreekVR: "Preserves" seems to refer to some decision you reached *while* building this, it says nothing to me as an end-user? What is this class *actually* for? When should I use it?

src/Services/Mapping/ApiMapperService.php
```diff
@@ -0,0 +1,736 @@
...
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefTypesCarrier;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefTypesDeliveryType;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefTypesDeliveryTypeV2;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentDefsDeliveryOptionsDeliveryNameV2;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentParametersCarrierName;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ShipmentResponsesDeliveryOptionsPackageTypeV2;
+use ReflectionClass;
+
+/**
+ * Maps carriers, delivery types and package types between delivery-options names, v1 ids and v2 names.
```
FreekVR: This class only maps carrier, delivery type and package type. Shipment options *do* have a mapping, but its in a `CapabilitiesMapper` model.

It'd be great if we could centralize the mappings in one place, so we don't have to hunt different files in different namespaces to find the one to update.

If that's not feasable, at least reference it here, so a user would know where to look for a shipment options map - and an implementer knows where that's maintained instead and why.

src/Services/Mapping/ApiMapperService.php
```diff
@@ -0,0 +1,736 @@
...
+
+    public const DOMAIN_CARRIER       = 'carrier';
+    public const DOMAIN_DELIVERY_TYPE = 'delivery_type';
+    public const DOMAIN_PACKAGE_TYPE  = 'package_type';
+
+    /**
+     * Generated enum classes supplying each column, keyed by domain.
+     *
+     * @var array<string, array<string, class-string>>
+     */
```
FreekVR, line 53: "domains" are referenced here, but what *is* a domain in this context. In other words, what do I need to know as a developer if I were to change, remove or add one of these domains. Do they map to something else? Or is it just an freely defined categorization?

src/Services/Mapping/ApiMapperService.php
```diff
@@ -0,0 +1,736 @@
...
+            self::COLUMN_ID          => RefShipmentPackageType::class,
+            self::COLUMN_V2_NAME     => RefShipmentPackageTypeV2::class,
+        ],
+    ];
+
+    /**
+     * Exceptions to shared-constant matching, keyed by domain, constant name and column.
+     *
+     * A value replaces the generated cell; null blocks its mapping. Aliases are input-only.
+     * Alias-only entries preserve older spellings after the generated canonical value changes.
```
FreekVR: Perhaps we can explain an alias in simpler terms?

src/Services/Mapping/ApiMapperService.php
```diff
@@ -0,0 +1,736 @@
...
+use ReflectionClass;
+
+/**
+ * Maps carriers, delivery types and package types between delivery-options names, v1 ids and v2 names.
+ *
+ * Generated constants are joined by name. Explicit overrides take precedence; name-to-name
+ * lookups may also use case conversion when the result exists in the target column.
+ * Unknown values return null. Invalid configuration or missing source classes fail at construction.
+ */
+final class ApiMapperService
```
FreekVR, line 34: This 750-line class gives me a bit of an "overengineered"-vibe. The public API seems good, but some of the internals seem like they could be made more simple. 

Are there any possibilities for simplications? Are we covering edge-cases that could be skipped given the audience for this class? Premature optimisations that cost more code than they cut in execution time?

FreekVR, line 34: This class seems to forget to document how to use it.

UPGRADE.md
```diff
@@ -399,3 +399,32 @@ The following classes and namespaces have been fully removed:
...
+
+```php
+use MyParcelNL\Sdk\Services\Mapping\ApiMapperService;
+
+$mapper = ApiMapperService::forCarrier();
+
+$mapper->idFromLegacyName('dhlforyou'); // 9
+$mapper->v2NameFromId(9);              // 'DHL_FOR_YOU'
+```
+
```
FreekVR, line 416: This is only documented in the upgrade guide, but where should we document it for new users?

FreekVR (changes requested):
API ziet er goed uit, nog een paar vraagjes voor verduidelijking/versimpleing

src/Model/Capabilities/CapabilitiesMapper.php
NabDevs: Renamed this to `v2NameToProperty`. It maps a v2 JSON name, such as `requiresSignature`, to the PHP property `requires_signature`.

src/Model/Capabilities/CapabilitiesMapper.php
NabDevs: Added `CapabilitiesRequest::getUnsupportedOptions()`. Callers can check which names will be omitted before sending, then decide whether to reject or change the request. Existing calls still skip these options.

src/Model/Capabilities/CapabilitiesMapper.php
NabDevs: Updated the names and comments to use "v2 JSON name" and "PHP property". The name conversion now lives in `ShipmentOptionMapper`.

src/Model/Shipment/Mapping/CarrierApiMapping.php
NabDevs: Updated the docblock to explain the conversion between v2 names and v1 IDs, and the exception for unknown values. It points to `ApiMapperService` for new code.

src/Services/Mapping/ApiMapperService.php
NabDevs: Moved the option mapping to `Services/Mapping/ShipmentOptionMapper`, next to this service. Options use model properties and have no IDs, so they have a separate mapper. Added a reference here.

src/Services/Mapping/ApiMapperService.php
NabDevs: A domain is a category, such as carrier or package type. The docblock now explains its three source definitions and how to add a category. Existing category keys must stay stable.

src/Services/Mapping/ApiMapperService.php
NabDevs: Added a simple example: `unfranked` is another accepted input for `letter`. The returned legacy name stays `letter`.

src/Services/Mapping/ApiMapperService.php
NabDevs: Reduced the class from 736 to 448 lines. Removed extra caches, alias indexes and runtime checks for fixed configuration. Tests now check that configuration. The public API stays the same.

NabDevs: Added a short usage example to the class docblock, with ID and name conversions.

UPGRADE.md
NabDevs: Added usage examples to the README, including the check for unsupported options. The upgrade guide now links to that section.

FreekVR (approved):

GravendeelJochem (approved):

PR #646, opened 2026-09-09
feat(api): add the ecommerce api factory

FreekVR:
Adds `EcommerceApiFactory`, which builds the generated e-commerce client with `DpopMiddleware` on it. It is the caller #645 shipped without. Stacked on #645.

Use it rather than constructing `DefaultApi` yourself: the generated client offers no way to put the middleware on the Guzzle client it sends with, and without it every call goes out unsigned and comes back 401.

The host comes from `ConnectConfig`. The spec carries no `servers` entry, so the generated default is `http://localhost` and the host is always set here. A `$host` argument overrides it for resource calls only, so a tunnel or a local mock stays possible while the connect calls keep going to the host `ConnectConfig` builds.

Resource calls carry the same `User-Agent` as the rest of the SDK, through `ConnectService::getUserAgentHeader()`, so an order push and a connect call read the same in a MyParcel log. It is fixed when the client is built, so a plugin registers itself with `setUserAgentForProposition()` first.

There is no API key version. The connect operations declare `security: []` and `POST /webhook/orders` declares nothing but `dpop`, so no operation in this spec could authenticate with a key. The design had one, written before the spec landed, when the assumption was the older public `POST /orders` with `apiKey` and `jwt` security. Say so and it is a few lines to add.

The test that matters is `testTheMiddlewareSitsInsideHttpErrorsSoItSeesA401`. Pushed outside `http_errors` instead of inside, a 401 would already be an exception before the retry could run, and #645's own tests would not notice because they build their own stack. It drives the stack this factory built with a mocked 401 then 202 and asserts both answers were used. Reaching the stack uses reflection on the generated client rather than an accessor, so a test needs no production surface.

Refs INT-1837
Refs INT-1838


src/Services/Ecommerce/EcommerceApiFactory.php
```diff
@@ -0,0 +1,93 @@
...
+        ?string $host = null,
+        ?string $userAgent = null
+    ): DefaultApi {
+        // The spec carries no servers entry, so the generated default is http://localhost. The host
+        // is always set here.
+        $resolvedHost = $host ?? $connect->getConfig()->getHost();
+
+        $config = new Configuration();
+        $config->setHost($resolvedHost);
+        $config->setUserAgent($userAgent ?? $connect->getUserAgentHeader());
```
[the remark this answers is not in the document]
FreekVR: Right on both counts, and reproduced:

```
empty host         host=''    ua='MyParcelNL-SDK/11.0.0-beta.33 php/8.4.20'
empty user agent   host='...' ua=''
```

An empty host sets both `Configuration::setHost('')` and Guzzle's `base_uri`, so every URL resolves against nothing. The empty user agent is the worse of the two: the generated client only writes the header `if ($this->config->getUserAgent())`, so `''` means no `User-Agent` is sent at all and the SDK's identifier disappears silently.

Both arguments now read `''` as not provided, matching `ShipmentApiFactory`, `WebhookApiFactory` and `IamApiFactory`. It is also a realistic input rather than hardening: `make($connect, $settings->get('host'))` with an unset setting hands over `''` rather than null.

Two tests, and reverting either one back to `??` fails them.

src/Services/Ecommerce/EcommerceApiFactory.php
```diff
@@ -0,0 +1,98 @@
...
+        // The spec carries no servers entry, so the generated default is http://localhost. The host
+        // is always set here.
+        $resolvedHost = '' !== (string) $host ? $host : $connect->getConfig()->getHost();
+
+        $config = new Configuration();
+        $config->setHost($resolvedHost);
+        $config->setUserAgent('' !== (string) $userAgent ? $userAgent : $connect->getUserAgentHeader());
+
+        return new DefaultApi(
+            new GuzzleClient([
```
NabDevs, line 78: The URL _can_ be redirected, which doesn't seem to be a likely scenario since it comes from our API, but tokens and/or order contents might leak this way. A simple reassuring fix would be to add `'allow_redirects' => false` to `EcommerceApiFactory::make()`.

NabDevs (changes requested):

src/Services/Ecommerce/EcommerceApiFactory.php
FreekVR: Added, with a test. It is worse than a maybe: Guzzle strips Authorization on a cross-origin redirect, but the redirected request comes back down the stack through DpopMiddleware, which signs it again for the new host. On a 307 or 308 the order body goes along too. I reproduced both before the change. With redirects off the 3xx surfaces as an ApiException and nothing goes to the other host.

NabDevs (approved):

joerivanveen (approved):

PR #649, opened 2026-09-10
feat(connect): add the connect proof of concept

FreekVR:
Adds a local page that runs the whole connect flow in a browser, so it can be driven by hand before any plugin exists. Stacked on #646.

`dev/connect-poc/README.md` has the prerequisites and how to run it. `/dev` is export-ignored, as `docker-compose.yml` already was, so none of it reaches a consumer's vendor directory.

`dev/connect-poc/index.php` is one page: buttons for connect, refresh, disconnect and uninstall, and a form that pushes a demo order with editable fields. It is ordered for reading, with everything a consumer implements first and the HTML and styling below a banner saying they are not part of using the SDK. Class names are written out in full, so it is clear which come from the generated client and which are hand written.

`dev/connect-poc/FileConnectStorage.php` is the example `ConnectStorageInterface`, and says why a JSON file is the wrong choice in a real integration: its writes take no lock, so two requests at once lose one of them.

Three changes to the SDK came out of running it:

- Added `ConnectService::CALLBACK_PATH`. The path was prose in three docblocks, so a consumer had to retype it into their router.
- `assertNonceMatches()` now separates a callback that carried no nonce from a shop with no flow waiting. It reported the second for both, which sent the reader to storage for a problem in the request. The existing test asserted only the exception type, and the branch for a missing stored nonce had no coverage.
- Narrowed `ConnectScope` and `ConnectPlatform` to what MyParcel accepts today: the `integration` scope and the `shopify` service. Measured against acceptance, where `generic` redirects to an Auth0 client that tenant does not know and `woo` answers 403 at `/connect/start`. The other entries are commented out rather than removed, pending confirmation from the team.

Verified on PHP 7.4: 659 tests, 29 skipped. The extra skip is `testTheHostLabelIsNotTheLowercasedPlatform`, which proves `WOOCOMMERCE` maps to `woo` and has nothing left to compare while only `shopify` is enabled.

Refs INT-1837
Refs INT-1838


NabDevs (approved):

dev/connect-poc/index.php
```diff
@@ -0,0 +1,762 @@
...
+ * 4. A route at your shop URL plus ConnectService::CALLBACK_PATH that finishes it. See
+ *    receiveCallback().
+ *
+ * Once connected, EcommerceApiFactory gives you an API client that signs every call. See pushOrder().
+ *
+ * The file is in two halves. Everything down to connectExceptionFacts() is what your integration
+ * does. Everything after that draws this page and is none of your concern.
+ *
+ * This file is not part of the SDK, and it is not a template to copy. It keeps state in a JSON file
+ * and has no login, so never put it on a host that is reachable from the internet.
```
joerivanveen, line 33: I agree with this prose, and see that this /dev directory is normally not included. However there might be scenarios where this is reachable, for instance on testing / staging servers where the SDK is cloned directly? And it might 'not work' by accident, also this comment does not prevent it from working. If other humans agree, I would like a 'guarantee' that this index will not do anything in a default scenario. For instance a required env var, if that is considered safe enough?

joerivanveen (changes requested):

dev/connect-poc/index.php
FreekVR: The /dev folder ships without a `.env` and the default fallback values would make it try to connect to acceptance, without a shop URL. The latter is mandatory for the script to even show anything.

Essentially it wouldn't be functional unless someone added the required .env values, too.

So the amount of things to go wrong are:

1) The SDK must be a full clone/download, not a composer install
2) The /dev folder must be reachable through http somehow
3) The /dev folder must have been made writable to the php process (for the storage to work)
4) The .env or global ENV vars must have been filled with at least the shop env and the production env toggle

FreekVR: We could add a big red banner on top to prevent any abuse, if you agree this works as a safe guard on top of these?

FreekVR: image

joerivanveen (approved):

